PR #310, opened 2022-10-24
304 replace json.stringify with safe stable stringify

AlexNi245:


AlexNi245:
During testing I noticed that profiles created prior this pr are no longer compatible because their signature was created using JSON.stringify instead of save-stringify. Hence i think it breaks the app for all existing users. We need to think about how this can be solved. [
Screenshot 2022-10-27 at 13 16 50
](url)

packages/lib/src/encryption/Encryption.ts
```diff
@@ -272,10 +270,10 @@ export function checkSignature(
     }
 }
 
-export function signWithSignatureKey(message: Message, keys: Keys): string {
+export function signWithSignatureKey(message: any, keys: Keys): string {
```
hai-ko, line 273: Was it necessary to change the type from `Message` to `any`?

packages/lib/src/encryption/Encryption.ts
```diff
@@ -272,10 +270,10 @@ export function checkSignature(
...
     return ethers.utils.hexlify(
         nacl.sign.detached(
-            ethers.utils.toUtf8Bytes(JSON.stringify(message)),
+            ethers.utils.toUtf8Bytes(stringify(message)!),
```
hai-ko, line 276: Can `stringify()` return something else than a string?

packages/lib/src/encryption/Encryption.ts
AlexNi245: I used this method also to sign an object from type Omit<Messaging.Message, 'signature'>. So it's no longer a message. The implementation expects an object of any type that can be stringified later.

packages/lib/src/encryption/Encryption.ts
AlexNi245: It can return string | null | undefined

hai-ko (approved):

PR #311, opened 2022-10-24
296 change to new message data structure

AlexNi245:


packages/lib/src/encryption/Encryption.ts
```diff
@@ -272,7 +272,7 @@ export function checkSignature(
     }
 }
 
-export function signWithSignatureKey(message: Message, keys: Keys): string {
+export function signWithSignatureKey(message: any, keys: Keys): string {
```
hai-ko, line 275: Why was the type changed from `Message` to `any`?

hai-ko (approved):

PR #312, opened 2022-10-25
290 add delivery service profile entry

### LINKED ISSUE #290: Add DeliveryServiceProfile entry
- [x] Change `UserProfile` to 
```
{
  // Key used to encrypt messages
  publicEncryptionKey: string,
  // Key used to sign messages
  publicSigningKey: string,
  // ENS name list of the delivery services e.g., delivery.dm3.eth
  deliveryServices: string[], 
  // URL pointing to the profile extension JSON file
  mutableProfileExtensionUrl: string
}
```
- [x] Add DeliveryServiceProfile
```
{
  // Key used to sign postmarks
  publicSigningKey: string,
  // Key used to encrypt delivery information
  publicEncryptionKey: string,
  // URL pointing to the delivery service instance
  url: string
}
```
- [x] Adjust lookup flow to get DeliveryServiceProfile. Use always `deliveryServices[0]` (#293 will take care of fallback feature)

AlexNi245:
This pr applies the new UserProfile interface described in the spec document to the dm3 repo. The following things are changed within this pr

-  Renamed the interfaces ProfileRegistryEntry to UserProfile and SignedProfielRegistryEntry to SignedUserProfile and changed the corresponding variables.
- Applied the new Data structure to the UserProfile as described in the spec. Hence the field publicKey is no longer part of the profile.
- Remove Public Key interface and extractPublicKey functions because they are no longer in need. 
- Set profile.deliveryService[0] as the default delivery service.
- Added DeliveryServiceProfile interface

closes #290


packages/lib/src/signin/Connect.ts
```diff
@@ -25,19 +22,19 @@ export async function connectAccount(
...
             throw Error('Profile signature is invalid');
         }
+        //Todo is this the right way to do it
+
         return {
             account,
-            existingAccount: profile?.profileRegistryEntry.publicKeys
-                ? true
-                : false,
+            existingAccount: profile !== undefined,
```
AlexNi245, line 37: Is this the proper way to check if an account already exists? Went with the check if the profile is undefined because according to the interface it's not possible that a key is undefined.

packages/lib/src/external-apis/BackendAPI.ts
```diff
@@ -33,7 +33,7 @@ export async function getChallenge(account: Account): Promise<string> {
     const checkedAccount = checkAccount(account);
     return (
         await axios.get(
-            (checkedAccount.profile.deliveryServiceUrl ?? '') +
+            (checkedAccount.profile.deliveryServices[0] ?? '') +
```
hai-ko, line 36: I think we should throw an error if `checkedAccount.profile.deliveryServices[0]` is null or undefined

packages/lib/src/external-apis/BackendAPI.ts
```diff
@@ -48,7 +48,7 @@ export async function getNewToken(
     const checkedAccount = checkAccount(account);
     return (
         await axios.post(
-            (checkedAccount.profile.deliveryServiceUrl ?? '') +
+            (checkedAccount.profile.deliveryServices[0] ?? '') +
```
hai-ko, line 51: same as above

packages/lib/src/external-apis/BackendAPI.ts
```diff
@@ -112,7 +112,7 @@ export async function syncAcknoledgment(
     const checkedAccount = checkAccount(connection.account);
     return axios.post(
         `${
-            (checkedAccount.profile.deliveryServiceUrl ?? '') + DELIVERY
+            (checkedAccount.profile.deliveryServices[0] ?? '') + DELIVERY
```
hai-ko, line 115: same as above

packages/lib/src/external-apis/BackendAPI.ts
```diff
@@ -148,7 +148,7 @@ export async function getNewMessages(
     return (
         await axios.get(
             `${
-                (checkedAccount.profile.deliveryServiceUrl ?? '') + DELIVERY
+                (checkedAccount.profile.deliveryServices[0] ?? '') + DELIVERY
```
hai-ko, line 151: same as above

packages/lib/src/external-apis/BackendAPI.ts
```diff
@@ -166,7 +166,7 @@ export async function getPendingConversations(
     return (
         await axios.post(
             `${
-                (checkedAccount.profile.deliveryServiceUrl ?? '') + DELIVERY
+                (checkedAccount.profile.deliveryServices[0] ?? '') + DELIVERY
```
hai-ko, line 169: same as above

packages/lib/src/external-apis/BackendAPI.ts
```diff
@@ -189,7 +189,7 @@ export async function getProfileRegistryEntryOffChain(
                 url
                     ? url
                     : `${
-                          (checkAccount(account).profile.deliveryServiceUrl ??
+                          (checkAccount(account).profile.deliveryServices[0] ??
```
hai-ko, line 192: same as above

packages/lib/src/signin/Connect.ts
hai-ko: Let us discuss this on a call

packages/react/src/Dm3.tsx
```diff
@@ -180,7 +180,7 @@ function dm3(props: dm3Props) {
             }
 
             const socket = socketIOClient(
-                state.connection.account.profile.deliveryServiceUrl,
+                state.connection.account.profile.deliveryServices[0],
```
hai-ko, line 183: `socketIOClient` needs a URL and `deliveryServices[0]` contains the ENS name of the delivery service. 
With the new structure, it is necessary to take the ENS name of the delivery service and get the URL from the delivery service record.

hai-ko (changes requested):

hai-ko (approved):

PR #313, opened 2022-10-26
301 delivery service  add postmark to envelop

### LINKED ISSUE #301: Delivery service: add `Postmark ` to Envelop
Postmark:
```
{
  // if unecrypted keccak256( safe-stable-stringify( EncryptionEnvelop.message ) ) 
  // if encrypted keccak256( EncryptionEnvelop.message ) 
  messageHash: string,
  // timestamp of when the delivery service received the message
  incommingTimestamp: number,
  // sign( keccak256( safe-stable-stringify( postmark_without_sig ) ) )
  signature: string
}
```

- [ ] Encrypt postmark with receiver public key

AlexNi245:
closes #301


packages/lib/src/delivery/Messages.ts
```diff
@@ -37,29 +47,82 @@ export async function getMessages(
...
+
+    const tokenIsValid = await checkToken(getSession, sender, token);
+    if (!tokenIsValid) {
+        //Token is invalid
         throw Error('Token check failed');
     }
+
+    const receiverSession = await getSession(receiver);
+    if (receiverSession === null) {
+        //TODO how should we handle this case?
```
AlexNi245, line 71: Is there a possiblity for the session being null ? If so how should we handle such case properly

packages/lib/src/delivery/Messages.ts
```diff
@@ -37,29 +47,82 @@ export async function getMessages(
...
+    );
+
+    //Encrypte the signed Postmark and return the ciphertext
+    const { ciphertext, nonce, version, ephemPublicKey } = encryptSafely({
+        publicKey: receiverEncryptionKey,
+        data: { ...postmarkWithoutSig, signature },
+        version: 'x25519-xsalsa20-poly1305',
+    });
+
+    return stringify({ nonce, version, ciphertext, ephemPublicKey })!;
```
AlexNi245, line 118: I just returned the stringified encrypted object. Should be further encoded i.E Base64 ?

packages/lib/src/messaging/Messaging.ts
```diff
@@ -154,19 +202,26 @@ export async function getMessages(
         (
             await getNewMessages(connection, userDb, contact)
         )
-            .filter((envelop) =>
-                envelop.deliveryServiceIncommingTimestamp ? true : false,
-            )
+            .filter((envelop) => {
```
AlexNi245, line 205: Why do we need to filter messages without an incommingTimestamp?

packages/lib/src/account/Account.test.ts
```diff
@@ -386,137 +342,121 @@ test('Should reject an invalid signature of a string', async () => {
...
+                    ? signedUserProfile
                     : undefined,
         ),
-    ).resolves.toStrictEqual(signedProfileRegistryEntry);
+    ).resolves.toStrictEqual(signedUserProfile);
 });
 
 test('publishProfileOnchain', async () => {
-    expect.assertions(2);
+    // expect.assertions(2);
```
hai-ko, line 459: Why was this line changed to a comment?

packages/lib/src/delivery/Messages.ts
hai-ko: I think we should discuss this on a call

packages/lib/src/delivery/Messages.ts
hai-ko: I think we should discuss this on a call

packages/lib/src/encryption/Encryption.ts
```diff
@@ -52,6 +53,7 @@ export function encrypt({
             try {
                 pubKeyUInt8Array = naclUtil.decodeBase64(publicKey);
             } catch (err) {
+                console.log(err);
```
hai-ko, line 56: `log()` from `shared/log` should be used instead of `console.log()`

packages/lib/src/messaging/Messaging.ts
hai-ko: I think it was because the specification requires a `deliveryServiceIncommingTimestamp,` but I'm not sure if that was the main reason.

packages/react/src/user-info/UserInfo.tsx
```diff
@@ -238,13 +238,13 @@ function UserInfo(props: UserInfoProps) {
                             <a
                                 className="text-decoration-none text-muted align-self-center"
                                 href={
-                                    props.account.profile.deliveryServiceUrl +
+                                    props.account.profile.deliveryServices[0] +
```
hai-ko, line 241: That won't work. The `deliveryServices[0]` is an ENS name. The app needs to resolve the ENS name and get the URL from the deliver service text record.

packages/react/src/user-info/UserInfo.tsx
```diff
@@ -238,13 +238,13 @@ function UserInfo(props: UserInfoProps) {
...
                                     '/profile/' +
                                     props.account.address
                                 }
                                 target="_blank"
                             >
-                                {props.account.profile.deliveryServiceUrl +
+                                {props.account.profile.deliveryServices[0] +
```
hai-ko, line 247: That won't work. The `deliveryServices[0]` is an ENS name. The app needs to resolve the ENS name and get the URL from the deliver service text record.

hai-ko (changes requested):

packages/lib/src/delivery/Messages.ts
hai-ko: Delivery service should only accept users with a session

packages/lib/src/delivery/Messages.ts
hai-ko: We let it unencoded

packages/lib/src/messaging/Messaging.ts
hai-ko: As discussed on a call: Remove filter

packages/backend/src/index.ts
```diff
@@ -53,6 +54,12 @@ let redisClient: undefined | Awaited<ReturnType<typeof createRedisClient>>;
     };
     app.locals.io = io;
 
+    app.locals.deliveryServiceProfile = {
+        publicSigningKey: process.env.PUBLIC_SIGNING_KEY,
```
hai-ko, line 58: We only need the private Key

packages/lib/src/delivery/Messages.ts
```diff
@@ -37,29 +47,85 @@ export async function getMessages(
...
+    const receiverSession = await getSession(receiver);
+    if (receiverSession === null) {
+        //TODO how should we handle this case?
+        throw Error('unknown session');
+    }
+
+    const receiverEncryptionKey =
+        receiverSession?.signedUserProfile.profile.publicEncryptionKey;
+
+    const deliveryServiceSigningKey = deliveryServiceProfile.publicSigningKey;
```
hai-ko, line 78: Must be private key

hai-ko (approved):

PR #335, opened 2022-11-09
318 validate schema for incoming requests

AlexNi245:


packages/lib/package.json
```diff
@@ -41,7 +42,7 @@
         "ts-json-schema-generator": "^0.98.0"
     },
     "scripts": {
-        "build:schema": "yarn ts-json-schema-generator -f tsconfig.json --path src/lib/Messaging.ts --type Message -o src/schema.json --no-type-check",
+        "build:schema": "yarn ts-json-schema-generator -f tsconfig.json --path src/lib/Messaging.ts --type Message -o src/schema.json --no-type-check && yarn ts-json-schema-generator -f tsconfig.json --path lib/src/delivery/Messages.ts --type Acknoledgment -o src/delivery/schema/Acknoledgment.schema.json --no-type-check && yarn ts-json-schema-generator -f tsconfig.json --path lib/src/messaging/Messaging.ts --type EncryptionEnvelop -o src/messaging/schema/EncryptionEnvelop.schema.json --no-type-check",
```
AlexNi245, line 45: Do you know if there is a better way to provide a script that's build all the schemas needed?

packages/lib/src/shared/validateSchema.ts
```diff
@@ -0,0 +1,13 @@
+import Ajv, { KeywordCxt } from 'ajv';
+import { ethers } from 'ethers';
+
+export function validateSchema(schema: any, data: any) {
+    const ajv = new Ajv();
+    try {
+        const validate = ajv.compile(schema);
+        return validate(data);
+    } catch (e) {
+        console.log(e);
```
hai-ko, line 10: log from lib should be used

hai-ko (changes requested):

PR #336, opened 2022-11-10
328 add deliver service properties

AlexNi245:


packages/backend/src/config/getDeliveryServiceProperties.ts
```diff
@@ -0,0 +1,29 @@
+import { parse } from 'yaml';
+import { existsSync, readFileSync } from 'fs';
+import * as Lib from 'dm3-lib/dist.backend';
+import { resolve } from 'path';
+
+const DEFAULT_CONFIG_FILE_PATH = resolve(__dirname, './../../src/config.yml');
```
hai-ko, line 6: I think we shouldn't point to something in `src`. `src` will not be available if we install a package with yarn or npm.

packages/backend/src/config/getDeliveryServiceProperties.ts
```diff
@@ -0,0 +1,29 @@
...
+    path: string = DEFAULT_CONFIG_FILE_PATH,
+    defaultDeliveryServiceProperties: Lib.delivery.DeliveryServiceProperties = DEFAULT_DELIVERY_SERVICE_PROPERTIES,
+): Lib.delivery.DeliveryServiceProperties {
+    if (!existsSync(path)) {
+        Lib.log('Config file not found. Default Config is used');
+        return defaultDeliveryServiceProperties;
+    }
+    const yamlString = readFileSync(path, { encoding: 'utf-8' });
+
+    const { messageTTL, sizeLimit } = {
```
hai-ko, line 23: Do we check the schema of the result object?

AlexNi245: Like if messageTTL i.E is a number?

packages/backend/src/config/getDeliveryServiceProperties.ts
AlexNi245: I replaced the src input with a relative one. In order to work the config.yml has to be copied after the build runs because TSC only handles .ts files. I added this to the package.json

hai-ko (approved):

PR #338, opened 2022-11-11
refactor: switch from xsalsa20 to chacha20

### LINKED ISSUE #294: Change default encryption to x25519-chacha20-poly1305

### LINKED ISSUE #339: Use padding for encryption

hai-ko:
closes #294
closes #339


packages/lib/src/delivery/Messages.ts
```diff
@@ -86,46 +88,39 @@ export async function incomingMessage(
...
-    return hexlify(toUtf8Bytes(encryptedPostmark));
+    };
 }
 
-const signPostmark = (p: Omit<Postmark, 'signature'>, signingKey: string) => {
+function signPostmark(
+    p: Omit<Postmark, 'signature'>,
+    signingKey: string,
+): Promise<string> {
+    log(`SING POST MARK USING ${signingKey}`);
```
AlexNi245, line 123: Do we want to keep this log statement? I think it's probably left from one of my PR's maybe we can remove It now.

packages/lib/src/account/Account.ts
```diff
@@ -31,22 +28,20 @@ import { UserProfileResolver } from './profileResolver/json/UserProfileResolver'
...
-}
-
 export interface UserProfile {
     publicEncryptionKey: string;
     publicSigningKey: string;
     deliveryServices: string[];
     mutableProfileExtensionUrl?: string;
 }
 
+export interface Keys {
```
AlexNi245, line 38: Wouldnt it makes sense to move this interface to the KeyCreation.ts file?

packages/lib/src/session/SignIn.ts
```diff
@@ -53,17 +67,19 @@ export async function signIn(
...
-                connection,
-                personalSign,
-                getSymmetricalKeyFromSignature,
-            );
+        const keys = await createKeys(
+            await createStorageKey(signedNonceMsg),
+            nonce,
+        );
 
+        if (!externalDataFile && !browserDataFile) {
```
AlexNi245, line 79: Is this just happen during development? Making a comment here would make sense to explain this code section.

packages/react/src/sign-in/Connectors.ts
```diff
@@ -169,9 +168,11 @@ export async function signIn(
                 break;
 
             case Lib.storage.StorageLocation.dm3Storage:
-                let authToken = (await localforage.getItem(
-                    'ENS_MAIL_AUTH_' + account.address,
-                )) as string;
+                let authToken = null;
+                // (await localforage.getItem(
```
AlexNi245, line 172: Why is this code commented out?

packages/react/src/utils/SigCheck.ts
```diff
@@ -0,0 +1,28 @@
...
+    accountAddress: string,
+    signature: string,
+): Promise<boolean> {
+    const isValid = await Lib.crypto.checkSignature(
+        publicSigningKey,
+        stringify(message)!,
+        signature,
+    );
+
+    if (!isValid) {
```
AlexNi245, line 16: Why not return false here already? After this there is no way the function can return true anymore.

AlexNi245 (changes requested):
See comments

packages/react/src/utils/SigCheck.ts
hai-ko: Yes, I simplified the function

packages/lib/src/session/SignIn.ts
hai-ko: There will be an issue addressing the sing in / auth.

AlexNi245 (approved):

PR #348, opened 2022-11-25
340 separate ds auth from sign in and store auth token

AlexNi245:


packages/lib/src/session/signInWithEtheruem.ts
```diff
@@ -0,0 +1,70 @@
...
+    const siweMessage = new SiweMessage({
+        //Get domain from widget i guess
+        domain: 'www.dm3.chat',
+        address: account,
+        statement,
+        uri: 'origin',
+        nonce: nonce.toString(),
+        version: '1',
+        chainId: 1,
+        issuedAt: new Date(12).getTime().toString(),
```
AlexNi245, line 26: How should we implement issuedAt ? If we are doing it properly it'll no longer possible to create the encryption key in an deterministic way

packages/lib/src/session/SignIn/signProfile.ts
```diff
@@ -0,0 +1,19 @@
...
+    personalSign: PersonalSign,
+    account: string,
+    stringifiedProfile: string,
+): Promise<string> {
+    try {
+        const profileCreationMessage =
+            getProfileCreationMessage(stringifiedProfile);
+        return await personalSign(provider, account, profileCreationMessage);
+    } catch (e) {
+        console.log(e);
```
hai-ko, line 16: remove `console.log`

packages/react/src/sign-in/Connectors.ts
```diff
@@ -138,143 +140,63 @@ export async function signIn(
...
-                type: ConnectionType.ChangeConnectionState,
-                payload: singInRequest.connectionState,
-            });
-        }
-        dispatch({
-            type: ConnectionType.ChangeConnectionState,
-            payload: singInRequest.connectionState,
-        });
-    }
+    console.log('Auth state', state.auth);
```
hai-ko, line 200: remove `console.log`

packages/react/src/ui-shared/RequestContacts.ts
```diff
@@ -37,20 +74,19 @@ export async function requestContacts(
         retrievedContacts = await Lib.account.getContacts(
             connection,
             userDb,
+            deliveryServiceToken,
             createEmptyConversationEntry,
         );
     }
+    console.log('GOT CONTRACT ', retrievedContacts);
```
hai-ko, line 81: remove `console.log`

hai-ko (changes requested):

hai-ko (approved):

PR #370, opened 2022-12-12
refactor: disallow console.log and .only tests

hai-ko:


packages/lib/src/account/Account.test.ts
```diff
@@ -532,7 +532,7 @@ test('publishProfileOnchain', async () => {
             return profile.signedUserProfile;
         },
     );
-
+    // eslint-disable-next-line no-console
```
AlexNi245, line 535: I think we should either use the logger here as well or remove this line completely

packages/lib/src/shared/log.ts
```diff
@@ -1,5 +1,6 @@
 export function log(text: string) {
     if (process.env.NODE_ENV !== 'production') {
+        // eslint-disable-next-line no-console
```
AlexNi245, line 3: eradicate

AlexNi245 (approved):

PR #381, opened 2022-12-20
feat: auth with sigining key

hai-ko:


packages/react/src/sign-in/getDatabase.ts
```diff
@@ -22,7 +22,12 @@ async function getExistingDatebase(
     state: GlobalState,
     dispatch: React.Dispatch<Actions>,
 ) {
-    const deliveryServiceToken = await Lib.session.reAuth(state.connection);
+    const keys = await Lib.session.createKeyPairsFromSig(state.connection, 0);
```
AlexNi245, line 25: Will the nonce always be 0?

AlexNi245 (commented):
See comment

packages/react/src/sign-in/getDatabase.ts
hai-ko: No, later on, there should be the possibility to change it but for now, would keep it at 0.
You need a new nonce to rotate the keys but that's a future feature.

AlexNi245 (approved):

PR #383, opened 2022-12-20
380 pending messages doesn't get a callback from lib

### LINKED ISSUE #380: `pendingMessages` doesn't get a callback from lib
#379 should be merged before working on this issue
BackendAPI.ts:
```
export async function createPendingEntry(
    ...
)
```
should look like:
```
export async function submitMessage(
   ...
    onSuccess: () => void,
    onError: () => void,
)
```

unskip test in `backend/src/messaging.test.ts` after fixing this

AlexNi245:
closes #380

### DIFF
--- a/packages/lib/src/messaging/Message.ts
+++ b/packages/lib/src/messaging/Message.ts
@@ -84,6 +84,8 @@ export async function submitMessage(
         deliveryServiceToken,
         message.metadata.from,
         message.metadata.to,
+        () => {},
+        () => {},
     );
     /**
      * The client can halt the delivry of a message if the receiver has no dm3-profile yet.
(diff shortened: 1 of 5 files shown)


packages/lib/src/messaging/Message.ts
```diff
@@ -84,6 +84,8 @@ export async function submitMessage(
         deliveryServiceToken,
         message.metadata.from,
         message.metadata.to,
+        () => {},
+        () => {},
```
AlexNi245, line 88: Do you think we should stop submitting the incoming msg and throw an error here if the pending entry could not be created?

hai-ko: I think we should ignore it wouldn't be a problem if the pending entry couldn't be created

hai-ko (approved):

PR #394, opened 2023-01-10
368 ccip lookup component

### LINKED ISSUE #386: CCIP Lookup Component
A user should be able to use dm3 without having a profile set onchain. Hence a service should be offer them a way to create temporarily profile using a subdomain of dm3.eth.
This service should become another package of the dm3 monorepo and should be hosted independently from other components.
It should also use signers that are deployed as separate web services that can be used to sign a profile. 
The following things should be built:

- [ ] Add an Endpoint that can be used to create an entire new users and save their profile 
- [ ] Sign a newly created profile with a private key similar as the delivery services does signing
- [ ] Add an Endpoint that returns a profile and the according signatures for a giving ens name

AlexNi245:
closes #386


packages/lib/src/offchainResolver/decodeCalldata.test.ts
```diff
@@ -0,0 +1,57 @@
...
+            dnsName('foo.dm3.eth'),
+            textData,
+        ]);
+
+        expect(() => decodeCalldata(calldata)).toThrowError(
+            "Namehash doesn't match",
+        );
+    });
+});
+function dnsName(name: string) {
```
hai-ko, line 33: Is this code copied from somewhere?

hai-ko: If yes we should check the license and reference it in a comment

packages/lib/src/offchainResolver/encodeUserProfile.test.ts
```diff
@@ -0,0 +1,72 @@
...
+        );
+
+        const [decodedProfile] = getResolverInterface().decodeFunctionResult(
+            'text',
+            encodedResult,
+        );
+        expect(JSON.parse(decodedProfile)).toStrictEqual(profile);
+    });
+});
+function dnsName(name: string) {
```
hai-ko, line 48: We should put this function in one place and import it if we need it.

hai-ko (changes requested):

hai-ko (approved):

Issue #401, opened 2023-01-12
Refactor old redis client


hai-ko:
- [ ] Use env for redis URL
- [ ] Remove second redis connection function

PR #406, opened 2023-01-16
feat: new sign-in flow

### LINKED ISSUE #400: New sign in flow
1. Connect metamask
2. check if address has a reverse record 
2.1. if yes, check if ENS name has a dm3 profile
2.2. if no, lookup offchain profile by address 
3. User has a profile?
3.1. Yes, sign in
3.2. No, create ccip profile and sign in

hai-ko:
closes #400


packages/backend/src/redis.ts
```diff
@@ -0,0 +1,78 @@
+import 'dotenv/config';
```
AlexNi245, line 1: I think this file was accidentally comitted since it was delete with #401

packages/lib/src/account/Account.ts
```diff
@@ -65,12 +63,29 @@ export function getProfileCreationMessage(stringifiedProfile: string) {
     return `Hearby your dm3 profile is linked with your Ethereum account\n\n ${stringifiedProfile}`;
 }
 
+/**
+ * calculate the namehash of a given ENS name
+ * @param account Account with ENS name
+ */
+export function getNamehash(account: Account): string {
```
AlexNi245, line 70: Haven't we decided to not do nameprep anymore because it causes errors if the ens-name is too long?

packages/lib/src/account/Account.ts
```diff
@@ -318,10 +332,21 @@ export function createHashUrlParam(profile: SignedUserProfile): string {
     return `dm3Hash=${sha256(stringify(profile))}`;
 }
 
+/**
+ * normalizes a namehash
+ * @param namehash the namehash that should be checked
+ */
+export function normalizeNamehash(namehash: string): string {
```
AlexNi245, line 339: Why do we need this function?

AlexNi245 (commented):
see comments

AlexNi245 (approved):

PR #417, opened 2023-02-06
feat: claim username

### LINKED ISSUE #416: Claim username

hai-ko:
closes #416


packages/backend/src/auth.ts
```diff
@@ -40,9 +40,10 @@ export default () => {
         '/:ensName',
         async (req: express.Request & { app: WithLocals }, res, next) => {
             try {
-                const ensName = Lib.account.normalizeEnsName(
+                const ensName = await req.app.locals.db.getId(
```
AlexNi245, line 43: maybe rename ensName to id or something since the variable is no longer containing an ens name

packages/backend/src/delivery.ts
```diff
@@ -36,10 +36,10 @@ export default () => {
         '/messages/:ensName/contact/:contactEnsName',
         async (req: express.Request & { app: WithLocals }, res, next) => {
             try {
-                const ensName = Lib.account.normalizeEnsName(
+                const ensName = await req.app.locals.db.getId(
                     req.params.ensName,
                 );
-                const contactEnsName = Lib.account.normalizeEnsName(
+                const contactEnsName = await req.app.locals.db.getId(
```
AlexNi245, line 42: same here

packages/backend/src/persistance/getId.ts
```diff
@@ -0,0 +1,13 @@
+import { Redis, RedisPrefix } from './getDatabase';
+import * as Lib from 'dm3-lib/dist.backend';
+
+export function getId(redis: Redis) {
```
AlexNi245, line 4: I think getId can be part of the Session package since it references the Session prefix

packages/backend/src/rpc/methods/handleResolveProfileExtension.ts
```diff
@@ -7,13 +7,12 @@ export function handleResolveProfileExtension(axios: Axios) {
...
         req: express.Request & { app: WithLocals },
         res: express.Response,
-        next: express.NextFunction,
     ) => {
         const {
             params: [ensName],
         } = req.body;
 
-        const normalizedEnsName = Lib.account.normalizeEnsName(ensName);
+        const normalizedEnsName = await req.app.locals.db.getId(ensName);
```
AlexNi245, line 15: rename var name

packages/lib/src/session/Connect.ts
```diff
@@ -26,7 +28,9 @@ export async function connectAccount(
         const address =
             preSetAccount ?? (await requestAccounts(connection.provider));
 
-        const ensName = await connection.provider.lookupAddress(address);
+        const ensName =
+            (await connection.provider.lookupAddress(address)) ??
```
AlexNi245, line 32: Maybe this can be extracted into a function

AlexNi245 (changes requested):
Pr looks great overall. I just left a few comments mostly regarding naming. One thing I noticed is that getId is often used within other persistence-related functions. 
We should keep in mind that this could probably cause circular dependencies. Maybe we can use the id as a constructor arg instead of the ensName. At least we should be aware of that issue.

packages/backend/src/auth.ts
hai-ko: I'll change the variable to `IdEnsName` and the function to `getIdEnsName` since it still contains an ENS name.
The `getIdEnsName` function returns the initial `0x123....addr.dm3.eth` ENS name which is internally used as the ID.

packages/backend/src/persistance/getId.ts
hai-ko: Yes, I changed it

hai-ko:
> Pr looks great overall. I just left a few comments mostly regarding naming. One thing I noticed is that getId is often used within other persistence-related functions. We should keep in mind that this could probably cause circular dependencies. Maybe we can use the id as a constructor arg instead of the ensName. At least we should be aware of that issue.

There is currently no Id besides the ENS name. The initial `addr.dm3.eth` ENS name is the Id. The idea is that if you pass `alice.user.dm3.eth` into the `getIdEnsName()` (prev. `getId()`) function you'll get `0x123...456.addr.dm3.eth` if you pass in `alice.eth` you'll also get `0x123...456.addr.dm3.eth` and if you pass in `0x123...456.addr.dm3.eth` it will also just return `0x123...456.addr.dm3.eth`.

I'm unsure if there is a risk of a circular dependency since `getIdEnsName()` doesn't depend on any other persistence-related function.

AlexNi245 (approved):
Pr looks great overall. I just left a few comments mostly regarding naming. One thing I noticed is that getId is often used within other persistence-related functions. 
We should keep in mind that this could probably cause circular dependencies. Maybe we can use the id as a constructor arg instead of the ensName. At least we should be aware of that issue.

PR #424, opened 2023-02-09
419 support top level ens name

### LINKED ISSUE #419: Support top level ENS name

AlexNi245:
closes #419

### DIFF
--- a/packages/react/src/domain-config/ConfigView.tsx
+++ b/packages/react/src/domain-config/ConfigView.tsx
@@ -2,6 +2,7 @@ import React, { useContext, useEffect, useState } from 'react';
 import { GlobalContext } from '../GlobalContextProvider';
 import * as Lib from 'dm3-lib';
 import { ConnectionType } from '../reducers/Connection';
+import { ethers } from 'ethers';
 
 function ConfigView() {
     const [addrEnsName, setAddrEnsName] = useState<string | undefined>();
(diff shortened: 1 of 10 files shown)


packages/react/src/domain-config/ConfigView.tsx
```diff
@@ -29,6 +31,50 @@ function ConfigView() {
...
+            setIsValidEnsName(false);
+            return;
+        }
+        const address = await state.connection.provider!.resolveName(ensName);
+
+        if (address === null) {
+            setIsValidEnsName(false);
+            return;
+        }
+        state;
```
hai-ko, line 65: What's that?

packages/react/src/domain-config/ConfigView.tsx
```diff
@@ -29,6 +31,50 @@ function ConfigView() {
...
+        event: React.FormEvent<HTMLInputElement>,
+    ) => {
+        const ensName = (event.target as any).value;
+        setEnsName(ensName);
+        const isValidEnsName = ethers.utils.isValidName(ensName);
+        if (!isValidEnsName) {
+            setIsValidEnsName(false);
+            return;
+        }
+        const address = await state.connection.provider!.resolveName(ensName);
```
hai-ko, line 59: Here you are getting the ETH record but I think you need to get the controller of the domain. Because the controller is the one that is able to set the records. Often the controller and the ETH record are the same but that is not always the case.

hai-ko (changes requested):

hai-ko (approved):

PR #425, opened 2023-02-10
423 initialize session from on chain profile

### LINKED ISSUE #423: initialize session from on-chain profile
The delivery service has to create a new session when a user with an existing on-chain profile tries to log in for the first time.
Currently, the app crashes because the delivery service tries to find the profile based on the alias which was never created in the first place. 

This error only happens if the user has created a dm3 profile on the chain without using the dm3 frontend.

AlexNi245:
closes #423

### DIFF
--- a/packages/lib/src/session/Connect.ts
+++ b/packages/lib/src/session/Connect.ts
@@ -1,20 +1,16 @@
+import axios from 'axios';
 import { ethers } from 'ethers';
 import { GlobalConf } from '..';
-import { GetUserProfile } from '../account';
+import { getUserProfile, GetUserProfile } from '../account';
 import { checkUserProfile, SignedUserProfile } from '../account/Account';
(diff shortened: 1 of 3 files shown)


packages/lib/src/session/Connect.ts
```diff
@@ -34,56 +28,110 @@ export async function connectAccount(
...
-        let account: string | undefined;
-
         const address =
             preSetAccount ?? (await requestAccounts(connection.provider));
 
-        const ensName = await findEnsNameToUse(connection.provider, address);
+        const ensName = await connection.provider.lookupAddress(address);
+        return ensName
+            ? await connectOnchainAccount(connection, ensName, address)
+            : await connectOffchainAccount(connection, address);
```
hai-ko, line 37: What's an offchain account?

hai-ko (approved):

PR #426, opened 2023-02-13
feat: new conf design

### LINKED ISSUE #420: Claim UI

hai-ko:
closes #420

### DIFF
--- a/packages/react/src/domain-config/Config.css
+++ b/packages/react/src/domain-config/Config.css
@@ -1,17 +1,19 @@
 .config-banner {
-    background-color: #ffc1073b;
+    background-color: #3A3C50;
     color: #ffffff;
     margin: 10px;
+    border-radius: 10px;
 
 }
 
 .config-btn {
-    color: #000;
-    background-color: #ffffff;
-    border-color: #ffffff
+    color: #ffffff;
+
+    border-color: #ffffff;
+
 }
 
-.config-btn:hover {
+/* .config-btn:hover {
     color: #000;
     background-color: #e9e9e9;
     border-color: #e9e9e9;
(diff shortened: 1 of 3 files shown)


packages/react/src/domain-config/Config.css
```diff
@@ -27,4 +29,64 @@
...
+    color: #ffffff;
+  
+}
+
+.user-info-text-muted {
+    color: #8f8f93 !important;
+}
+
+.user-info-text-highlight {
+    color: #ffffff !important;
```
AlexNi245, line 50: Why so many !importants?

packages/react/src/domain-config/ConfigView.tsx
```diff
@@ -4,13 +4,39 @@ import * as Lib from 'dm3-lib';
...
+import Icon from '../ui-shared/Icon';
 
 function ConfigView() {
     const [addrEnsName, setAddrEnsName] = useState<string | undefined>();
     const [dm3UserEnsName, setDm3UserEnsName] = useState<string | undefined>();
+    const [existingDm3UserEnsName, setExistingDm3UserEnsName] = useState<
+        string | undefined
+    >();
+
+    const [loading, setLoading] = useState<
```
AlexNi245, line 16: What is the purpose of this var?

packages/react/src/domain-config/ConfigView.tsx
```diff
@@ -79,34 +105,42 @@ function ConfigView() {
...
-        window.location.reload();
+            dispatch({
+                type: ConnectionType.ChangeAccount,
+                payload: {
+                    ...state.connection.account!,
+                    ensName:
+                        dm3UserEnsName + Lib.GlobalConf.USER_ENS_SUBDOMAIN(),
+                },
+            });
+            window.location.reload();
```
AlexNi245, line 139: Why do we have to reload here?

AlexNi245 (commented):
see comments

packages/react/src/domain-config/ConfigView.tsx
hai-ko: This should be removed in a separate issue. For now, it was easier to re-sign-in instead of making all the changes in the background

packages/react/src/domain-config/ConfigView.tsx
hai-ko: This var stores the name of the component which is loading.

hai-ko: It is used to indicate loading and blocks the other button during loading

packages/react/src/domain-config/Config.css
hai-ko: Not sure if they are necessary but the complete style needs to be refactored in a separate issue.

packages/react/src/domain-config/ConfigView.tsx
AlexNi245: I'd expect loading to be bool when it is solely used to indicate whether a component is loading. Is there a better naming for it?

AlexNi245 (approved):

packages/react/src/domain-config/ConfigView.tsx
hai-ko: It can't be bool because it indicates what is loaded

PR #428, opened 2023-02-28
build: beta2

hai-ko:


packages/react/src/contacts/Contacts.tsx
```diff
@@ -22,16 +22,10 @@ function Contacts(props: ContactsProps) {
...
             props.getContacts(state.connection);
         }
-    }, [state.userDb?.conversationsCount]);
-
-    useEffect(() => {
-        if (state.userDb?.conversations) {
-            props.getContacts(state.connection);
-        }
-    }, []);
+    }, [state.userDb?.conversations]);
```
AlexNi245, line 28: I think state.userDb?.conversationsCount has to be part of the dependency array otherwise it'll cause unusedDependecy warning

PR #480, opened 2023-04-17
476 add fetch all messages to ds

AlexNi245:


packages/backend/src/persistance/messages/getIncomingMessages.ts
```diff
@@ -0,0 +1,24 @@
+import { Redis, RedisPrefix } from '../getDatabase';
+import { getMessages } from './getMessages';
+import * as Lib from 'dm3-lib/dist.backend';
+
+export function getIncomingMessages(redis: Redis) {
```
AlexNi245, line 5: Maybe you find a better name for this?

packages/backend/src/persistance/messages/getIncomingMessages.ts
```diff
@@ -0,0 +1,24 @@
...
+    return async (
+        ensName: string,
+        limit: number,
+    ): Promise<Lib.messaging.EncryptionEnvelop[]> => {
+        const conversationIds = await redis.zRange(
+            RedisPrefix.IncomingConversations + ensName,
+            0,
```
AlexNi245, line 12: We're fetching just the first 100 conversations. I think that's sufficient. Otherwise we've to implement how we can implement pagination without knowing how much conversations we've in total

packages/lib/packages/delivery/src/Messages.ts
```diff
@@ -143,6 +143,11 @@ export async function incomingMessage(
 
     const envelopWithPostmark: EncryptionEnvelop = {
         ...envelop,
+        metadata: {
```
AlexNi245, line 146: Currently, we're writing encrypted DeliveryInformation in the db. In order to make this feature work we've to write the encrypted struct instead. Do you think that has implications regarding security etc. ??

hai-ko: no I think that's fine. If the DS gets hacked the Delivery Information are leaked but they are already contained in the unencrypted `conversationId`

packages/backend/src/persistance/messages/getIncomingMessages.ts
hai-ko: I think it's fine

PR #482, opened 2023-04-24
Repair lib imports

AlexNi245:


package.json
```diff
@@ -30,14 +39,14 @@
...
-        "start": "cd packages/backend && yarn start",
+        "build": "yarn workspaces run build",
+        "start": "cd ./packages/backend && yarn start",
         "resolver": "cd packages/offchain-resolver && yarn start",
         "lint": "eslint '*/**/*.{js,ts,tsx}' --fix",
         "format": "prettier --write '*/**/*.{js,ts,tsx,json}'",
         "format:check": "prettier --check '*/**/*.{js,ts,tsx,json}'",
         "prepare": "husky install",
-        "test": "npm run docker:up && ./test-all.sh"
+        "test": "./test-all.sh"
```
hai-ko, line 49: This should be replaced with something like `yarn workspaces run test`
test-all.sh can then be deleted.

packages/billboard-client/.env
```diff
@@ -0,0 +1,11 @@
+NODE_ENV=development
```
hai-ko, line 1: This file shouldn't be committed.
Maybe add `**.env` to root .gitignore

hai-ko (changes requested):

hai-ko (approved):

PR #485, opened 2023-04-25
feat: message proxy

hai-ko:


packages/lib/messaging/src/MessageProxy.ts
```diff
@@ -0,0 +1,60 @@
...
+            }
+        } catch (e) {
+            log(
+                `Failed to send message using: ${dsName} (${
+                    (e as Error).message
+                })`,
+            );
+        }
+    }
+    if (!success) {
```
AlexNi245, line 57: I don't think you need that if statement because the function never will reach this part of the function when at least one promise is returned successfully. 
Hence it might be sufficient just to throw here and remove the "success" var.

AlexNi245 (changes requested):
see comment

AlexNi245 (approved):

PR #486, opened 2023-04-26
484 create lib package for billboard api

AlexNi245:


packages/lib/billboard-api/src/client/axios/methods/getActiveSenders.ts
```diff
@@ -0,0 +1,24 @@
+import { Axios } from 'axios';
+import { log } from 'dm3-lib-shared';
+
+export function getActiveViewers(axios: Axios) {
+    return async (idBillboard: string): Promise<number | null> => {
```
hai-ko, line 5: These inner functions share a lot of the same  lines. I think they could be replaced with:
```
async function jsonRpcRequest<T>(axios:Axios, method: string, parms: any[], url: string): Promise<T | null> => {

        const body = {
            jsonrpc: '2.0',
            method,
            params,
        };

        const { data } = await axios.post(url, { data: body });

        const { error, result } = data;

        if (error) {
            log(error.message);
            return null;
        }

        return result;
    };
```
Usage example:
```
export function getActiveViewers(axios: Axios) {
   return () => jsonRpcRequest<number>(axios, 'dm3_billboard_countActiveViewers', [idBillboard], '/rpc')
}
```

hai-ko (approved):

PR #488, opened 2023-05-04
Integration tests

hai-ko:


packages/integration-tests/src/flow.test.ts
```diff
@@ -0,0 +1,287 @@
...
+
+    const pushMessage = (ensName: string, envelop: EncryptionEnvelop) => {
+        pushReg[ensName](envelop);
+    };
+
+    beforeEach(() => {
+        pushReg = {};
+        textRecordMockReg = {};
+        httpPostRequests = {};
+        aliceWallet = new ethers.Wallet(
```
AlexNi245, line 65: maybe it could make sense to use ethers.Wallet.createRandom() here

packages/lib/shared/src/jsonDataUri.ts
```diff
@@ -0,0 +1,5 @@
+import { stringify } from './stringify';
+
+export function createJsonDataUri(data: any): string {
```
AlexNi245, line 3: I think it'll make sense to name the file the name as the function including if just one function is exported

AlexNi245 (changes requested):
see comments

packages/lib/profile/src/Profile.ts
```diff
@@ -216,13 +215,18 @@ export async function createProfile(
     accountAddress: string,
     deiveryServiceNames: string[],
```
AlexNi245, line 216: typo

packages/lib/profile/src/Profile.ts
```diff
@@ -186,21 +187,19 @@ export function isSameEnsName(
 async function createKeyPairsFromSig(
     provider: ethers.providers.JsonRpcProvider,
```
AlexNi245, line 188: I think it'll make more sense to pass a signer instead of a provider here. That would make the function much easier to use by components not based in the FE

packages/integration-tests/src/flow.test.ts
```diff
@@ -0,0 +1,287 @@
+import { createDeliveryServiceProfile, createProfile } from 'dm3-lib-profile';
```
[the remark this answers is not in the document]
hai-ko: Yes, a clearer structure makes sense. I'll create a lib sub-dir. I can't put it into Profile or Message because this test concerns the complete lib

AlexNi245 (approved):

PR #492, opened 2023-05-12
feat: profile wildcards for resolver

hai-ko:


packages/offchain-resolver-backend/src/utils/intercept.ts
```diff
@@ -0,0 +1,45 @@
...
+
+        return { interceptor, higherLevelDomain };
+    }
+}
+
+// intercepts a text record query if the higher level domain name is registered
+// in the interceptor env variable
+export function interceptTextRecord(ensName: string, textRecordName: string) {
+    try {
+        const interceptorContainer = getInterceptor(ensName);
```
AlexNi245, line 22: Maybe destructuring the interceptor container might increase readability
`const {interceptor,higherLevelDomain} = getInterceptor(ensName)`

AlexNi245 (changes requested):
see comments

packages/offchain-resolver-backend/src/utils/intercept.ts
hai-ko: I couldn't destruct it since it may return undefined but I moved the `higherLevelDomain`-Check to `getInterceptor` so it should be more readable now

packages/offchain-resolver-backend/src/utils/intercept.ts
```diff
@@ -0,0 +1,45 @@
+import { log } from 'dm3-lib-shared';
```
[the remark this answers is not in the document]
hai-ko: Agreed, I moved it

AlexNi245 (approved):

PR #506, opened 2023-05-30
feat: siwe

hai-ko:


packages/offchain-resolver-backend/src/http/profile.ts
```diff
@@ -6,6 +6,7 @@ import {
 import { ethers } from 'ethers';
 import express from 'express';
 import { WithLocals } from './types';
+import { SiweMessage } from 'siwe';
 
 //The test msg should just be the sg of an ethereum address
 const MSG_START = 0;
```
AlexNi245, line 12: I think we can get rid of this since we're using the siwe class to parse the message

packages/offchain-resolver-backend/src/http/profile.ts
```diff
@@ -23,17 +24,16 @@ export function profile(web3Provider: ethers.providers.BaseProvider) {
                 const { signedUserProfile, siweMessage, siweSig, hotAddr } =
                     req.body;
 
+                const siwe = new SiweMessage(JSON.parse(siweMessage));
```
AlexNi245, line 27: It would be handy if we can add logs for siweMessage, to enhance debugging

AlexNi245: also if the app breaks during production it likely occurs here. Maybe we can add some additional monitoring

AlexNi245 (changes requested):
see comments

packages/offchain-resolver-backend/src/http/profile.ts
AlexNi245: Also, it might sense to wrap the Json.parse in a try-catch because we're not sanitizing the input at all and the server would crash with 500 if siweMessage is not al valid JSON string

AlexNi245 (approved):

PR #515, opened 2023-06-06
Billboard client load test

AlexNi245:


packages/billboard-client/package.json
```diff
@@ -29,7 +29,8 @@
     "scripts": {
         "docker:up": "docker-compose up -d",
         "start": "yarn ts-node --transpile-only  ./src/index.ts",
-        "test": "npm run docker:up && jest --coverage --runInBand --forceExit",
+        "test": "npm run docker:up && jest  --coverage --runInBand --forceExit --modulePathIgnorePatterns='<rootDir>/test/load'",
```
hai-ko, line 32: Couldn't we use `yarn` instead of `npm`

packages/billboard-client/package.json
```diff
@@ -29,7 +29,8 @@
...
+        "test:load": "npm run docker:up && jest /test/load/LoadTest.test.ts --coverage --runInBand --forceExit",
```
hai-ko, line 33: Couldn't we use `yarn` instead of `npm`

PR #537, opened 2023-07-13
refactor: delivery and offchain api

hai-ko:


packages/lib/delivery-api/src/auth-http.ts
```diff
@@ -0,0 +1,61 @@
...
+import {
+    Account,
+    getDeliveryServiceClient,
+    normalizeEnsName,
+} from 'dm3-lib-profile';
+
+import { ethers } from 'ethers';
+import { checkAccount } from './utils';
+
+const AUTH_SERVICE_PATH = process.env.REACT_APP_BACKEND + '/auth';
```
AlexNi245, line 11: Is this supposed to be used within both FE and BE ? If it might make sense to rename the REACT_APP_BACKEND var

AlexNi245 (approved):

PR #551, opened 2023-07-27
544 signin page

Bhupesh-mfsi:
Task : https://github.com/corpus-io/dm3/issues/544


packages/messenger-widget/src/components/SignIn/bl.tsx
hai-ko, line 1: What's the meaning of the file name?

Bhupesh-mfsi: **contexts/shared.ts** changed to  **contexts/Shared.ts**

The lower case changed to upper case

Bhupesh-mfsi: **bl.tsx** refers to **bussiness logic**, If there are too many logics and functions  then those are included in this file.

hai-ko: Ok makes sense

hai-ko (approved):

Issue #555, opened 2023-07-31
Create UI for action item


Bhupesh-mfsi:
## Acceptance Criteria

1. Create a UI for dropdown menu containing **Show details** & **Hide contact** option.
2. On click of action item(three dots), drop down should open containing **Show details** & **Hide contact** option.
3. On click of **Show details**, the contact information page of the contact should open.
4. On click of **Hide contact**, the contact should hide and not be visible in the contact list and default DM3 chat screen should open.

PR #566, opened 2023-08-07
555 contacts action

Bhupesh-mfsi:
Task : #555 
Figma : https://www.figma.com/file/SqAsFElnxNzP69pic9RQah/dm3-embedded?type=design&node-id=85-2636&mode=design&t=lXGNgL8G7B3a8MDI-0


packages/messenger-widget/src/components/Contacts/bl.ts
```diff
@@ -18,11 +18,14 @@ export function MouseOver(
     className: string,
```
AlexNi245, line 18: Please give the file a proper name

packages/messenger-widget/src/components/Contacts/bl.ts
```diff
@@ -86,7 +89,7 @@ export const onContactSelected = (
     // add click css
     event.currentTarget.classList.add(classTwo);
     const actionItem: any = document.getElementById('contact-' + index);
-    actionItem.style.display = 'none';
+    // actionItem.style.display = 'none';
```
AlexNi245, line 92: Do you intend to keep using that var ? Otherwiese please remove the comment

packages/messenger-widget/src/components/Contacts/bl.ts
```diff
@@ -192,9 +195,13 @@ export const setContactSelectedFromCache = (
     state: GlobalState,
     cacheContacts: ContactPreview[],
 ): number | null => {
-    const name = state.accounts.selectedContact?.account.ensName;
+    const key =
+        state.accounts.selectedContact?.account.profile?.publicEncryptionKey;
     for (let index = 0; index < cacheContacts.length; index++) {
```
AlexNi245, line 200: If there is no good reason to use for() please use a higher order function. I think .filter() would work for that case

AlexNi245 (changes requested):
see comments

packages/messenger-widget/src/components/Contacts/bl.ts
Bhupesh-mfsi: Oh yes, I will remove that

packages/messenger-widget/src/components/Contacts/bl.ts
Bhupesh-mfsi: Ok, will replace for loop with findIndex high order function as I needed index here.

packages/messenger-widget/src/components/Contacts/bl.ts
AlexNi245: okay if you've agreed on that with @hai-ko it's fine for me

AlexNi245 (approved):

Issue #559, opened 2023-07-31
Allow user to add conversation


Bhupesh-mfsi:
## Acceptance Criteria

1. Create a UI for popup modal to add conversation.
2. Show error message if wrong name is entered.

PR #571, opened 2023-08-11
559 add conversation

Bhupesh-mfsi:
Task : #559
Figma : https://www.figma.com/file/SqAsFElnxNzP69pic9RQah/dm3-embedded?type=design&node-id=63-3832&mode=design&t=QLrywQmbb3J9QKH5-0


packages/messenger-widget/src/adapters/contacts.ts
```diff
@@ -215,14 +215,22 @@ export async function getContacts(
...
+                try {
+                    profile = await getUserProfile(
+                        connection.provider!,
+                        ensName,
+                    );
+                    return {
+                        ensName,
+                        profile: profile,
+                    };
+                } catch (error) {
```
AlexNi245, line 228: Can the code proceed with the profile being undefined ?  If not might it make more sense to handle the error instead of just returning undefined

packages/messenger-widget/src/components/AddConversation/AddConversation.tsx
```diff
@@ -0,0 +1,164 @@
...
+    const [name, setName] = useState<string>('');
+    const [showError, setShowError] = useState<boolean>(false);
+    const [errorMsg, setErrorMsg] = useState<string>('');
+    const [inputClass, setInputClass] = useState<string>(INPUT_FIELD_CLASS);
+
+    // handles new contact submission
+    const submit = (e: React.FormEvent) => {
+        e.preventDefault();
+        setName(name.trim());
+        if (name.replace(SPACE_REGEX, '').length) {
```
AlexNi245, line 21: What is this check for ?

packages/messenger-widget/src/components/AddConversation/bl.ts
```diff
@@ -0,0 +1,193 @@
...
+};
+
+// updates hidden contact states
+const updateHiddenContact = (
+    state: GlobalState,
+    dispatch: React.Dispatch<Actions>,
+    normalizedAccountName: string,
+    aka: string | undefined,
+) => {
+    if (!aka) {
```
AlexNi245, line 149: what is aka?

packages/messenger-widget/src/components/Contacts/bl.ts
```diff
@@ -195,13 +195,30 @@ const getMessagesFromUser = (
 
 export const setContactSelectedFromCache = (
```
AlexNi245, line 196: The function returns a number instead of a contract. So either rename it to setContactIndexSelectedFromCache or make it return a contact

packages/messenger-widget/src/components/Contacts/bl.ts
```diff
@@ -195,13 +195,30 @@ const getMessagesFromUser = (
...
+        payload: 'Fetching contacts...',
+    });
+    startLoader();
+
     const key =
         state.accounts.selectedContact?.account.profile?.publicEncryptionKey;
-    const index = cacheContacts.findIndex(
+    const name = state.accounts.selectedContact?.account.ensName;
+
+    let index = cacheContacts.findIndex(
```
AlexNi245, line 212: use const instead of let

AlexNi245 (changes requested):
see comments

packages/messenger-widget/src/components/AddConversation/bl.ts
Bhupesh-mfsi: It is a string to handle hidden contacts which is used in **DM3 storage library**. This property is used in the UserDB named interface.

packages/messenger-widget/src/adapters/contacts.ts
Bhupesh-mfsi: Yes, it can proceed. When we add new contacts they might have not configured the profile, but will have ENS name and address that we need to show in contact list. Although chatting won't be allowed with those users until they configure profile, but that will be a check in chatting screen. But here simply it returns undefined showcasing that user has not configured profile.

AlexNi245 (approved):

Issue #560, opened 2023-07-31
Allow user to add preferences


Bhupesh-mfsi:
Create a UI for popup modal of **Preferences**

PR #572, opened 2023-08-14
560 preferences

Bhupesh-mfsi:
Task : #560 
Figma : https://www.figma.com/file/SqAsFElnxNzP69pic9RQah/dm3-embedded?type=design&node-id=85-2774&mode=design&t=90fbgUBL1je56g4V-0


packages/messenger-widget/src/components/Preferences/bl.tsx
```diff
@@ -0,0 +1,15 @@
+// method to open the preferences modal
+export const openPreferencesModal = () => {
+    let modal: HTMLElement = document.getElementById(
```
AlexNi245, line 3: const pls :)

AlexNi245 (changes requested):
see comments

packages/messenger-widget/src/components/Preferences/bl.tsx
Bhupesh-mfsi: The value of the variable is changed in next line that is why I didn't make it const. Shall I make it to const?

AlexNi245: I'd use const because the value of a property within the object is changed instead the object itself.

Bhupesh-mfsi: Ok, modified

AlexNi245 (approved):

PR #583, opened 2023-08-23
feat: remove alias

hai-ko:


packages/offchain-resolver/src/persistance/schema.prisma
```diff
@@ -13,6 +13,14 @@ model ProfileContainer {
   profile Json
   ensName String @unique
   address String @unique
+  aliases Alias[]
 
 }
 
+model Alias {
+    id Int @id @default(autoincrement())
```
AlexNi245, line 21: Do we need an additional ID when the namehash is part of the profile container ?

AlexNi245 (approved):

packages/offchain-resolver/src/persistance/schema.prisma
hai-ko: I checked it it seems that the `String` wasn't the issue.

AlexNi245 (approved):

PR #584, opened 2023-08-24
Notifications

AlexNi245:


packages/backend/src/config/getDeliveryServiceProperties.ts
```diff
@@ -31,7 +32,14 @@ export function getDeliveryServiceProperties(
         //The interface DeliveryServiceProperties requires all properties to be non-null. But since we are accepting a partially filled config.yml we are overwriting the required fields so basically no property is required at all. This can be done because every missing property is replaced by a default property
         {
             ...schema.DeliveryServiceProperties,
-            required: [],
+            definitions: {
```
hai-ko, line 35: Normally the auto-generated is based on a TS type. Couldn't we do this here too?

packages/lib/delivery/src/Messages.test.ts
```diff
@@ -271,7 +295,7 @@ describe('Messages', () => {
             }
         });
         //TODO remove skip once spam-filter is implemented
```
hai-ko, line 297: Comment can be removed

packages/lib/delivery/src/notifications/channels/Email.ts
```diff
@@ -0,0 +1,49 @@
...
+export function Email(config: EmailNotificationServerConfig) {
+    const send = async (
+        mailConfig: EmailNotificationUserConfig,
+        deliveryInformation: DeliveryInformation,
+    ) => {
+        nodemailer.createTestAccount();
+        const transport = nodemailer.createTransport(new SMTPTransport(config));
+
+        try {
+            const t = await transport.sendMail({
```
hai-ko, line 36: The variable `t` isn't used.

packages/backend/src/persistance/notification/addUsersNotificationChannel.ts
```diff
@@ -0,0 +1,73 @@
...
+        const isValid = validateSchema(schema.NotifcationChannel, channel);
+
+        if (!isValid) {
+            throw Error('Invalid NotificationChannel');
+        }
+        //Secondly we check if the type specific config matches
+        switch (channel.type) {
+            case NotifificationChannelType.EMAIL: {
+                const isValidConfig = validateSchema(
+                    {
```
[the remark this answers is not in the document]
AlexNi245: Yes specifying the format email not only checks whether the input is a string but also ensures a valid email was passed. To my knowledge, though the typegenerator is not capable of taking a format into account so i've added it

packages/backend/src/config/getDeliveryServiceProperties.ts
AlexNi245: The schema is already auto-generated via TS. However, the TS type requires every value to be required. This will be the case after the config reader has finished, but when reading the config every type is optional because when not specified additionally it'll be substituted with the default value for that property.

hai-ko (approved):

PR #613, opened 2023-09-12
Dm3 cli

AlexNi245:


packages/cli/README.md
```diff
@@ -0,0 +1 @@
+# cli
```
hai-ko, line 1: I think one example would make sense

PR #615, opened 2023-09-13
New UI

hai-ko:


.yarn/versions/3dae3775.yml
```diff
@@ -1,16 +1,16 @@
 undecided:
```
AlexNi245, line 1: Do you know if we need that file? If so is there a more apparent name ?

hai-ko: This is auto generated from yarn. 

https://yarnpkg.com/getting-started/qa

> .yarn/versions is used by the [version plugin](https://yarnpkg.com/features/release-workflow) to store the package release definitions. **You will want to keep it within your repository.**

AlexNi245 (approved):

PR #621, opened 2023-09-21
New storage

AlexNi245:


packages/lib/storage/src/new/MessageStorage.ts
```diff
@@ -0,0 +1,399 @@
...
+        );
+        //Add the chunk to the conversation
+        this.addLeaf(emptyChunk);
+        //finally the chunk that'll contain the message has been created and sits at the end of the list
+        await emptyChunk.add(message);
+        await this.save();
+    }
+}
+
+class Chunk extends Node {
```
hai-ko, line 370: I think we need different Chunk types. Currently we only have chunks which can store messages but e.g. we also need a chunk that can store metadata related to a conversation (e.g. is the conversation archived).

packages/lib/storage/src/new/MessageStorage.ts
```diff
@@ -0,0 +1,399 @@
...
+        //Return the messages
+        return chunk.getLeafs<Envelop>();
+    };
+
+    return { addMessage, getMessages, getConversations, _tree: root };
+};
+
+type Leaf = Node | Envelop;
+
+type ChunkIdentifier = {
```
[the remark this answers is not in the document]
hai-ko: Talk with Steffen regarding specification of Id

packages/lib/storage/src/new/index.ts
```diff
@@ -0,0 +1,143 @@
...
+            );
+            await db.keyValueStoreLocal.write(
+                messageChunkContainer.conversationManifest.key,
+                messageChunkContainer.conversationManifest,
+            );
+            if (
+                messageChunkContainer.messageChunk.envelops.length ===
+                MAX_MESSAGES_PER_CHUNK
+            ) {
+                await db.keyValueStoreLocal.write(
```
AlexNi245, line 80: Isnt that the same operation as in line 67

packages/lib/storage/src/new/index.ts
```diff
@@ -0,0 +1,143 @@
...
+    accountEnsName: string,
+    sign: (data: string) => Promise<string>,
+    options?: Partial<{
+        readStrategy: ReadStrategy;
+        keyValueStoreRemote?: KeyValueStore;
+        encryption?: Encryption;
+    }>,
+): StorageAPI {
+    const keyValueStoreLocal = createKeyValueStore();
+    const db: Db = {
```
AlexNi245, line 113: I think this should be DI too

packages/lib/storage/src/new/index.ts
```diff
@@ -0,0 +1,143 @@
...
+        readStrategy: ReadStrategy.LocalFirst,
+        accountEnsName,
+        keyValueStoreLocal,
+        sign,
+        // If we read from remote because a chunk is not available, we need to update local storage
+        updateLocalStorageOnRemoteRead: options?.keyValueStoreRemote
+            ? async <T extends Chunk>(key: string, value: T) => {
+                  await keyValueStoreLocal.write(key, value);
+              }
+            : undefined,
```
AlexNi245, line 123: would it make more sense to return a function here to prevent null pointer exceptions when trying to call a function on undefined?

packages/lib/storage/src/new/read.ts
```diff
@@ -0,0 +1,224 @@
...
+/**
+ * This function retrieves a specific chunk of messages for a given conversation from the storage.
+ *
+ * @param {Db} db - The database object, which contains the account's ENS name and the sign function.
+ * @param {string} contactEnsName - The ENS name of the contact.
+ * @param {number} page - The page number of the message chunk to retrieve.
+ * @returns {Promise<MessageChunk | undefined>} A promise that resolves to the message chunk
+ *  or undefined if the chunk does not exist.
+ */
+export async function getMessageChunk(
```
AlexNi245, line 27: I think it would make sense to have either a class or a function that receives the db object as a constructor param instead of drilling it down every function arg

const storageReader = new StorageReader(db) / getStorageReader(db)
strageReader.read(conversation,page)

packages/lib/storage/src/new/write.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    // calculate the key for the conversation list chunk and creates a new conversation list
+    // with the new conversation added
+    const key = await getConversationListKey(db, targetChunkIndex);
+
+    return {
+        conversationList: {
+            key,
+            conversationList: conversationListChunk
+                ? [...conversationListChunk.conversationList, contactEnsName]
```
AlexNi245, line 59: that would put the contactEnsName at the end of the conversationList. Could it make sene to insert it at the beginning so latest conversations are fetched at first when paginated?

packages/lib/storage/src/new/write.test.ts
```diff
@@ -0,0 +1,37 @@
...
+import { db, makeEnvelop, testEnvelop } from './testHelper';
+
+describe('write', () => {
+    describe('addConversation', () => {
+        it('should add a conversation to the database', async () => {
+            const contactEnsName = 'c2';
+
+            expect(await addConversation(contactEnsName, db)).toEqual({
+                conversationList: {
+                    key: '0x9c7f05ff5af8632a1b4f9112e3451e265696646ad70187d083282b3d8886be88',
```
AlexNi245, line 11: can we generate the keys dynamically so it is understandable what actually is tested here

packages/lib/storage/src/new/read.ts
```diff
@@ -0,0 +1,224 @@
...
+        return localChunk as T;
+    }
+
+    if (db.keyValueStoreRemote) {
+        // Only try to read from remote if the local chunk does not exist
+        const remoteChunk: T | undefined = await db.keyValueStoreRemote.read(
+            key,
+        );
+
+        if (remoteChunk) {
```
AlexNi245, line 165: if(remoteChunk && remoteFetchCb)

packages/lib/storage/src/new/read.ts
```diff
@@ -0,0 +1,224 @@
...
+}
+
+async function remoteFirstRead<T extends Chunk>(
+    key: string,
+    db: Db,
+    remoteFetchCb: RemoteFetchCb | undefined,
+) {
+    if (db.keyValueStoreRemote) {
+        const remoteChunk = await db.keyValueStoreRemote.read(key);
+        if (remoteChunk) {
```
AlexNi245, line 184: if(remoteChunk && remoteFetchCb)

packages/lib/storage/src/new/read.ts
```diff
@@ -0,0 +1,224 @@
...
+            if (remoteFetchCb) {
+                remoteFetchCb(key, remoteChunk);
+            }
+            // Return the remote chunk if it exists
+            return remoteChunk as T;
+        }
+    }
+
+    const localChunk = await db.keyValueStoreLocal.read(key);
```
AlexNi245, line 193: could duplication can be reduced by having two functions readChunkFromLocal and readChunkFromRemote that are beeing called in remoteFirstRead and localFirstRead

PR #660, opened 2023-11-21
Billboard staging

hai-ko:


packages/billboard-widget/src/hooks/useAuth.ts
```diff
@@ -88,7 +88,7 @@ export const useAuth = (
             wallet.address,
         );
 
-        const ensName = `${clientProps.siweAddress}.user.ethprague.dm3.eth`;
+        const ensName = `${clientProps.siweAddress}.bb-user.dm3.eth`;
```
AlexNi245, line 91: It think in one of the branches we had for istanbul this was already part of the client props obj

packages/cli/profile/index.ts
```diff
@@ -0,0 +1,66 @@
...
+
+const newProfile = async (program: Command) => {
+    const { profilePk, deliveryService } = program.opts();
+    if (!deliveryService) {
+        program.error(
+            'error: option --deliveryService <deliveryService> argument missing',
+        );
+    }
+    const storageKeyCreationMessage = getStorageKeyCreationMessage(
+        '0xca8f04fdc80d659997f69b02',
```
AlexNi245, line 22: make this value a constant to explain better what it stands for

AlexNi245 (changes requested):
see comments

PR #685, opened 2023-12-04
Prod deploy

hai-ko:


packages/offchain-resolver/src/persistance/migrations/20230825081456_init/migration.sql
```diff
@@ -0,0 +1,37 @@
+-- CreateTable
```
AlexNi245, line 1: If we're starting with a fresh db we can delete all previous migrations and just have a initial one

AlexNi245 (changes requested):
see comments

PR #687, opened 2023-12-05
Work--toplevel-alias-2

steffenkux:
The SC for the TLAR with tests


packages/toplevel-alias/contracts/Lock.sol
```diff
@@ -0,0 +1,34 @@
+// SPDX-License-Identifier: UNLICENSED
```
AlexNi245, line 1: Remove dummy contract

packages/toplevel-alias/contracts/TopLevelAliasRegistry.sol
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * @title TopLevelAliasRegistry
+ * @dev A contract for managing a registry of aliases mapped to names.
+ * This contract allows the owner to set aliases for given names, ensuring aliases end with '.eth',
+ * have at least three characters before '.eth', do not exceed a maximum length, and are not empty.
+ */
+contract TopLevelAliasRegistry is Ownable {
+    // Public mapping of names to aliases
+    mapping(string => string) public aliases;
```
AlexNi245, line 14: I think it could make sense to use the chainId as a key here. That makes it easier to query the contract and reduces storage costs.

packages/toplevel-alias/contracts/TopLevelAliasRegistry.sol
```diff
@@ -0,0 +1,64 @@
...
+     * Only the owner of the contract can call this function.
+     * Validates that the name is not empty, the alias length is within limits, and ends with '.eth'.
+     *
+     * @param _name The name to map the alias to. Must not be empty.
+     * @param _alias The alias to be set for the given name. Must meet the criteria.
+     */
+    function setAlias(string memory _name, string memory _alias) public onlyOwner {
+        require(bytes(_name).length > 0, "Name cannot be empty");
+        require(bytes(_alias).length >= 7 && bytes(_alias).length <= MAX_ALIAS_LENGTH, "Alias length is invalid");
+        require(_endsWith(_alias, ".eth"), "Alias must end with '.eth'");
```
AlexNi245, line 40: I wonder if it would be sufficient to store only the alias label without the eth postfix onchain since its a top level alias that is always settled at ens. 

To compute the name of a domain 
1. Name template is already given `alice.${TLD}.eth`
2. Get chain from injected provider
3. use chainId of provider to fetch alias from contract
4. supply alias to template -> `alice.gno.eth`

AlexNi245: We could simply check if only one label exists in the provided string via ENS stringUtils lib https://github.com/ensdomains/ens-contracts/blob/5ae0cd7500215fa0c6c8baa7491a4bda1b1634dd/contracts/wrapper/NameWrapper.sol#L394

packages/toplevel-alias/contracts/TopLevelAliasRegistry.sol
AlexNi245: https://docs.ens.domains/ens-improvement-proposals/ensip-9-multichain-address-resolution

packages/toplevel-alias/contracts/_TestTopLevelAliasRegistry.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
```
AlexNi245, line 2: migrate to hardhat test

packages/toplevel-alias/test/Lock.ts
```diff
@@ -0,0 +1,128 @@
+import {
+    time,
+    loadFixture,
+} from '@nomicfoundation/hardhat-toolbox/network-helpers';
+import { anyValue } from '@nomicfoundation/hardhat-chai-matchers/withArgs';
+import { expect } from 'chai';
+import { ethers } from 'hardhat';
+
+describe('Lock', function () {
```
AlexNi245, line 9: remove

AlexNi245 (approved):

PR #701, opened 2024-01-02
feat: new storage backend

hai-ko:


packages/backend/src/persistance/storage/getUserStorage.ts
```diff
@@ -3,9 +3,14 @@ import { UserStorage } from 'dm3-lib-storage';
 import { getIdEnsName } from '../getIdEnsName';
 
 export function getUserStorage(redis: Redis) {
```
AlexNi245, line 5: I wonder if we could use a more concise naming for that method since it is used to retrive a certain storage slot key. 
Maybe something like getStorageSlot / readSlot

packages/lib/storage/src/new/RemoteInterface.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @returns {KeyValueStore} A key-value store object.
+ */
+export function createRemoteKeyValueStoreApi(
+    baseUrl: string,
+    encryption?: Encryption,
+): KeyValueStore {
+    return {
+        write: async (key: string, value: Chunk) => {
+            const valueToStore = encryption
+                ? await encryption.encrypt(stringify(value))
```
AlexNi245, line 21: I think we could get rid of the if statement by using a default parameter that either is the dm3 encryption or a function that jus returns the input.

class Dm3Encryption implements Encryption {
...
}

export function createRemoteKeyValueStoreApi(
    baseUrl: string,
    encryption: Encryption = dm3Encryption,
): KeyValueStore {
   read: async <T extends Chunk>(key: string) => {
            const result = (await axios.get(`${baseUrl}/${key}`)).data;
            if (!result) {
                return undefined;
            }
            const processedResult =
           await encryption.decrypt(result.toString())

            return JSON.parse(processedResult) as T;
        },
    };

AlexNi245 (changes requested):
see comments

PR #717, opened 2024-01-17
702 dialog

Bhupesh-mfsi:
notifications UI


packages/messenger-widget/src/components/ConfigureProfile/bl.tsx
```diff
@@ -226,21 +220,28 @@ export const namingServices = [
...
+                chain = Number(process.env.REACT_APP_CHAIN_ID);
+            } else {
+                chain = namingServices[0].chainId;
+            }
+            return <ConfigureEnsProfile chainToConnect={chain} />;
         case NAME_SERVICES.GENOME:
-            return namingServices[1].component;
+            chain = namingServices[1].chainId;
+            return <ConfigureGenomeProfile chainToConnect={chain} />;
+        default:
```
AlexNi245, line 243: This will always fails since there is no chain 1

packages/messenger-widget/src/components/ConfigureProfile/bl.tsx
```diff
@@ -226,21 +220,28 @@ export const namingServices = [
...
     },
     {
         name: NAME_SERVICES.GENOME,
         chainId: 100,
-        component: <ConfigureGenomeProfile />,
     },
 ];
 
 export const fetchComponent = (name: string) => {
+    let chain: number = 1;
```
AlexNi245, line 231: remove chain var. Instead it within the context
`
      case NAME_SERVICES.ENS:
            if (process.env.REACT_APP_CHAIN_ID === '5') {
                return <ConfigureEnsProfile chainToConnect={5} />;
            } 
            return <ConfigureEnsProfile chainToConnect={1} />;
      case NAME_SERVICES.GENOME:
            const genomeChainId = namingServices[1].chainId;
            return <ConfigureGenomeProfile chainToConnect={genomeChainId} />;

`

packages/messenger-widget/src/components/Preferences/DM3Profile.tsx
```diff
@@ -0,0 +1,20 @@
+import { ConfigureDM3Profile } from '../ConfigureProfile/ConfigureProfile';
+import { ConfigureProfileContextProvider } from '../ConfigureProfile/context/ConfigureProfileContext';
+import { Heading } from './Heading';
+
+export function DM3Profile() {
```
AlexNi245, line 5: I like the structure of the Preferences component. Continue keeping components short and breaking them into small pieces. That helps to keep the code maintainable.

packages/messenger-widget/src/components/Preferences/Notification/DeleteIcon.tsx
```diff
@@ -0,0 +1,25 @@
+import './Notification.css';
+import deleteIcon from '../../../assets/images/delete.svg';
+import { VerificationMethod } from './bl';
+
+interface IDeleteIcon {
+    data: string;
+    type: VerificationMethod;
+    deleteNotification: Function;
```
AlexNi245, line 8: As long as there is no further specification on how to Notifications are implemented using the callback function to mock the behaviour is fine.
However once we have built the actual implementation that should be moved towards a hook that is handling the logic behind Notifications.

packages/messenger-widget/src/components/Preferences/Notification/bl.ts
```diff
@@ -0,0 +1,187 @@
...
+
+export const INPUT_FIELD_CLASS =
+    'notification-input-field font-weight-400 font-size-14 border-radius-6 w-100 line-height-24';
+
+export enum VerificationMethod {
+    Email = 'Email',
+    Telephone = 'Telephone',
+}
+
+export const getVerficationModalContent = (
```
AlexNi245, line 15: Instead of keeping every function in  bl.ts please move the Logic towards its hook that encapsulates everything related to Notifications, That hook should expose an interface for every functionality required by the Components in the Notification directory. 
The logic of the different methods ie mail, push notification etc. can be moved to their own functions that can be tested independently 

hooks
-Notifications
--useNotifications // provides an interface to all methods
--Methods
---email
---sms
---push
---...

packages/messenger-widget/src/components/Preferences/Notification/bl.ts
```diff
@@ -0,0 +1,187 @@
...
+    inputData: string,
+    setErrorMsg: Function,
+    setShowError: Function,
+): boolean => {
+    /* eslint-disable max-len */
+    const emailRegex =
+        /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
+    const phoneRegex = /^\(?([0-9]{3})\)?[-. ]?([0-9]{3})[-. ]?([0-9]{4})$/;
+
+    if (type === VerificationMethod.Email) {
```
AlexNi245, line 91: don't use nested if instead use && 
if (type === VerificationMethod.Email && !emailRegex.test(inputData)){
...
}

packages/messenger-widget/src/contexts/Accounts.ts
```diff
@@ -43,6 +43,14 @@ export function accountsReducer(state: Accounts, action: AccountsActions) {
                 contacts: action.payload,
             };
 
+        case AccountsType.Reset:
```
AlexNi245, line 46: What is the purpose of this functions ?

packages/messenger-widget/src/contexts/Shared.ts
```diff
@@ -23,10 +23,12 @@ const getMainnetProvider = () => {
 export const initialState: GlobalState = {
     connection: {
         defaultServiceUrl: process.env.REACT_APP_BACKEND as string,
+        reset: undefined,
```
AlexNi245, line 26: Why is there a reset property on every object?

packages/messenger-widget/src/hooks/auth/useAuth.ts
```diff
@@ -1,21 +1,31 @@
...
+    UiStateType,
+    UiViewStateType,
+} from '../../utils/enum-type-utils';
 import { useTopLevelAlias } from '../topLevelAlias/useTopLevelAlias';
 
 export const useAuth = (onStorageSet: (userDb: UserDB) => void) => {
     const { resolveAliasToTLD } = useTopLevelAlias();
     const { data: walletClient } = useWalletClient();
     const mainnetProvider = useMainnetProvider();
+    const { dispatch } = useContext(GlobalContext);
```
AlexNi245, line 28: dispatch is about to be removed and will be replaced by hooks over time. useAuth was the first hook and more about to follow after the state refactoring has been finished.
Hence please don't use dispatch anymore for any new developments. If you still have to perform an action within the legacy state management use a callback function similar to onStorageSet

AlexNi245 (changes requested):
See comments

packages/messenger-widget/src/contexts/Accounts.ts
Bhupesh-mfsi: In the current implementation, when the account changes the context API states do not get reset. So I added a reset action for every context states so that on account change we can use this action and reset entire context storage and it only resolved the issue of contacts and accounts header displaying the wrong details.

packages/messenger-widget/src/contexts/Shared.ts
Bhupesh-mfsi: In the current implementation, when the account changes the context API states do not get reset. So I added a reset action for every context states so that on account change we can use this action and reset entire context storage.

packages/messenger-widget/src/hooks/auth/useAuth.ts
Bhupesh-mfsi: Ok

packages/messenger-widget/src/components/ConfigureProfile/bl.tsx
Bhupesh-mfsi:
```
export const namingServices = [
    {
        name: NAME_SERVICES.ENS,
        chainId: 1,
    },
    {
        name: NAME_SERVICES.GENOME,
        chainId: 100,
    },
];

export const fetchComponent = (name: string) => {
    let chain: number = 1;
    switch (name) {
        case NAME_SERVICES.ENS:
            if (process.env.REACT_APP_CHAIN_ID === '5') {
                chain = Number(process.env.REACT_APP_CHAIN_ID);
            } else {
                chain = namingServices[0].chainId;
            }
            return <ConfigureEnsProfile chainToConnect={chain} />;
        case NAME_SERVICES.GENOME:
            chain = namingServices[1].chainId;
            return <ConfigureGenomeProfile chainToConnect={chain} />;
        default:
            return <ConfigureEnsProfile chainToConnect={chain} />;
    }
};```

Here as there are two chains, Ethereum mainnet with chain ID 1 and Gnosis with chain id 100. If user is connected to goerli for development it returns ENS that is  namingServices[0] with chain ID 5. If connected to mainnet then ENS with chainID 1 and if Gnosis then returns with chainID 100.

What's wrong here?

Bhupesh-mfsi: Here as there are two chains, Ethereum mainnet with chain ID 1 and Gnosis with chain id 100. If user is connected to goerli for development it returns ENS that is  namingServices[0] with chain ID 5. If connected to mainnet then ENS with chainID 1 and if Gnosis then returns with chainID 100.

What's wrong here?

AlexNi245: The default case will be never reached because it'll always return earlier.  Is case the chain is mainnet it'll return at the end of the if statement. The additional return and the let var are adding unneccessary complexity to the function and makes it harder to understand it

packages/messenger-widget/src/contexts/Shared.ts
AlexNi245: Yeah but what is the purpose of having a reset prop on every reducer ?

Bhupesh-mfsi: We need to reset data of every reducer as new account is connected. So it is resetted to default state. Since every reducer is created separately so I had to reset all. As we move further and remove legacy state managment code, this will also get removed

AlexNi245: I understand that we need to reset but why do we need a reset propery that seems to be always set to undefined ?

Bhupesh-mfsi: If you see the reducer code, for reset action all reducers are set to their initial state. undefined is used everywhere just to remain consistent. In reset actually we don't need to set specific value so undefined is taken. We just need trigerring of this action no matter what the payload is, so undefined is set.

AlexNi245: But what would happen if the reset param will be removed entirely ? The reset action would then return the initialState

AlexNi245 (approved):

PR #719, opened 2024-01-18
adding npm publish script

steffenkux:
/scripts/publish_npm.sh is a script running recursively over all packages. If a package contains a file `.export-npm`, it will published to npm

added yarn command `publish:npm`, too.


packages/backend/package.json
AlexNi245, line 1: Why do we need the .export-npm file ?

scripts/publish_npm.sh
```diff
@@ -0,0 +1,71 @@
...
+    cd "$dir"
+
+    # Read the package name from package.json
+    package_name=$(jq -r '.name' package.json)
+
+    # Get the current version from npm
+    current_npm_version=$(npm view "$package_name" version 2>/dev/null)
+
+    # Check if new version is greater than current version
+    if [ -n "$new_version" ] && version_gt "$new_version" "$current_npm_version"; then
```
AlexNi245, line 51: Do we have to always release every package when using this script ?

scripts/publish_npm.sh
AlexNi245, line 1: The executable for that package should be added into the script section of the main package.json, so it can be executed by simply using yarn deploy-npm

AlexNi245 (changes requested):
see comments

scripts/publish_npm.sh
steffenkux: For now, the script is for releasing all. We may extend it late for more detailed features.

scripts/publish_npm.sh
steffenkux: Habe ich ja drin:
        "publish:npm": "scripts/publish_npm.sh"

packages/backend/package.json
steffenkux: This is intended as flag that this package is to be published. Not all packages are published to npm.

steffenkux: Another option would be to maintain a list.

PR #733, opened 2024-01-24
730 email notify server config

Bhupesh-mfsi:
Email notification server configuration

### DIFF
--- a/packages/backend/src/config/getDeliveryServiceProperties.ts
+++ b/packages/backend/src/config/getDeliveryServiceProperties.ts
@@ -10,6 +10,11 @@ const DEFAULT_DELIVERY_SERVICE_PROPERTIES: DeliveryServiceProperties = {
     //100Kb
     sizeLimit: 100000,
     notificationChannel: [],
+    smtpHost: '',
+    smtpPort: 0,
+    smtpEmail: '',
+    smtpUsername: '',
+    smtpPassword: '',
 };
 
 export function getDeliveryServiceProperties(
--- a/packages/backend/src/config/getDeliveryServiceProperties.ts
+++ b/packages/backend/src/config/getDeliveryServiceProperties.ts
@@ -10,11 +10,11 @@ const DEFAULT_DELIVERY_SERVICE_PROPERTIES: DeliveryServiceProperties = {
     //100Kb
     sizeLimit: 100000,
     notificationChannel: [],
-    smtpHost: '',
-    smtpPort: 0,
-    smtpEmail: '',
-    smtpUsername: '',
-    smtpPassword: '',
+    smtpHost: process.env.SMTP_HOST as string,
+    smtpPort: Number(process.env.SMTP_PORT),
+    smtpEmail: process.env.SMTP_EMAIL as string,
+    smtpUsername: process.env.SMTP_USERNAME as string,
+    smtpPassword: process.env.SMTP_PASSWORD as string,
 };
 
 export function getDeliveryServiceProperties(
--- a/packages/backend/src/config/getDeliveryServiceProperties.ts
+++ b/packages/backend/src/config/getDeliveryServiceProperties.ts
@@ -10,11 +10,11 @@ const DEFAULT_DELIVERY_SERVICE_PROPERTIES: DeliveryServiceProperties = {
     //100Kb
     sizeLimit: 100000,
     notificationChannel: [],
-    smtpHost: process.env.SMTP_HOST as string,
-    smtpPort: Number(process.env.SMTP_PORT),
-    smtpEmail: process.env.SMTP_EMAIL as string,
-    smtpUsername: process.env.SMTP_USERNAME as string,
-    smtpPassword: process.env.SMTP_PASSWORD as string,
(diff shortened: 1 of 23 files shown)


packages/backend/src/config/getDeliveryServiceProperties.ts
```diff
@@ -10,6 +10,11 @@ const DEFAULT_DELIVERY_SERVICE_PROPERTIES: DeliveryServiceProperties = {
     //100Kb
     sizeLimit: 100000,
     notificationChannel: [],
+    smtpHost: process.env.SMTP_HOST as string,
```
AlexNi245, line 13: If we use the config file the config should bot be loaded from process.env

packages/lib/delivery/src/notifications/broker/NotificationBroker.ts
```diff
@@ -55,10 +56,12 @@ export const _setupNotficationBroker = (
  */
 export const NotificationBroker = (
     notificationChannel: NotificationChannel[],
+    notificationType: NotificationType,
 ): INotificationBroker => {
     const channels = notificationChannel.map((channel) => {
         switch (channel.type) {
             case NotificationChannelType.EMAIL:
+                channel.config.notificationType = notificationType;
```
AlexNi245, line 64: Why do we assign the notificationType to the config ?

packages/lib/delivery/src/notifications/utils.ts
```diff
@@ -0,0 +1,38 @@
+import { DeliveryInformation } from '@dm3-org/dm3-lib-messaging';
```
AlexNi245, line 1: It think it would be more concise to name the file fetchEmailTemplates or similiar instead of utils.ts

AlexNi245 (changes requested):
see comments

packages/lib/delivery/src/notifications/utils.ts
Bhupesh-mfsi: I thought to have utility functions like fetchEmailTemplates() and generateOtp() function in this file that will be used for notification purpose. So that it would not end up in creating multiple files for small functions.

packages/backend/src/config/getDeliveryServiceProperties.ts
Bhupesh-mfsi: Then in default value shall I put it empty string?

packages/lib/delivery/src/notifications/broker/NotificationBroker.ts
Bhupesh-mfsi: It is added so that everytime notificationBroker( ) is called we can fetch which kind of email is to be send like : OTP or NEW MESSAGE. And then from here we pass notificationType to _setupNotficationBroker() which calls the send() method to send email. And in that send() we have that config and accordingly we can decide whether its a OTP email or a NEW MESSAGE EMAIL so we can fetch that email template.

packages/backend/src/config/getDeliveryServiceProperties.ts
AlexNi245: Yes

packages/lib/delivery/src/notifications/broker/NotificationBroker.ts
AlexNi245: I think the channel.config.notificationType will never be referenced because its only accessible during the map call. Once the function has returned the Garbage collector will remove it

packages/lib/delivery/src/notifications/utils.ts
AlexNi245: I prefer to have two different files because you can find them better via file search and they are more visible in the file tree. Whereas utils.ts barely gives any information about the functions it contains.
However, it's not a dealbreaker for me. If you prefer this kind of naming pattern it's fine for me too.

Bhupesh-mfsi: No, I will follow the same structure that you have been following so there will be consistency. I will create separate files for both functions.

packages/lib/delivery/src/notifications/broker/NotificationBroker.ts
Bhupesh-mfsi: map function will return then values to channels and then we are calling     

**return _setupNotficationBroker(channels);**

It should work

AlexNi245: I see NotificationType is the type of Mail that is about to be send (OTP, NEW_MESSAGE) right ? In that case please add a comment explaining what the difference between channel and type is. Because it is a bit confusing without knowing about it

AlexNi245: And please update the TS doc of NotificationBroker so it contains the right params

Bhupesh-mfsi: Added comments & updated TS doc

AlexNi245 (approved):

PR #742, opened 2024-01-30
739 notification apis

Bhupesh-mfsi:


packages/backend/src/notifications.test.ts
```diff
@@ -20,7 +21,7 @@ const keysA = {
 
 describe('Notifications', () => {
     describe('get NotificationChannels', () => {
-        it('Returns 200 and an empty array when the user has no chanels set up', async () => {
+        it('Returns 500 as global notification is turned off', async () => {
```
AlexNi245, line 24: I think it would be better to return an empty array here aswell, since using the BE without any notification channel is completely fine

packages/backend/src/notifications.test.ts
```diff
@@ -90,18 +177,104 @@ describe('Notifications', () => {
...
+                .post(`/bob.eth`)
+                .set({
+                    authorization: `Bearer ${token}`,
+                })
+                .send({
+                    recipientValue: 'bob@gamil.com',
+                    notificationChannelType: NotificationChannelType.EMAIL,
+                });
+
+            expect(status).toBe(500);
```
AlexNi245, line 238: Please use Code 404 instead along with an error message that the particular message is not supported yet

packages/backend/src/notifications.ts
```diff
@@ -60,21 +60,41 @@ export default () => {
...
+                });
+            }
+
+            // Fetch global notification data of user from database
+            const globalNotification =
+                await req.app.locals.db.getGlobalNotification(account);
+
+            // Throw error if global notification is turned off
+            if (!globalNotification.isEnabled) {
+                throw new Error('Global notifications is off');
```
AlexNi245, line 90: 400 with error message

packages/backend/src/validation/notification/regexValidation.ts
```diff
@@ -0,0 +1,21 @@
...
+    notificationChannelType: string,
+    recipientValue: string,
+): boolean => {
+    switch (notificationChannelType) {
+        case NotificationChannelType.EMAIL:
+            // checks regex pattern of email ID
+            const patternCheck = String(recipientValue)
+                .toLowerCase()
+                .match(
+                    /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/,
```
AlexNi245, line 15: please make it a constant

packages/backend/src/validation/notification/regexValidation.ts
```diff
@@ -0,0 +1,21 @@
...
+                );
+            return patternCheck ? true : false;
```
AlexNi245, line 17: simply return patternCheck

AlexNi245 (changes requested):
see comments

packages/backend/src/validation/notification/regexValidation.ts
Bhupesh-mfsi: patternCheck either return an array or null, so for better readability boolean is returned

AlexNi245 (approved):

PR #747, opened 2024-02-07
741 otp

Bhupesh-mfsi:
Email verification OTP


packages/backend/src/config/getDeliveryServiceProperties.test.ts
```diff
@@ -89,23 +74,28 @@ describe('ReadDeliveryServiceProperties', () => {
...
                 },
             ],
-            smtpHost: 'smtp.host',
-            smtpPort: 587,
-            smtpEmail: 'dm3@gmail.com',
-            smtpUsername: 'dm3@gmail.com',
-            smtpPassword: 'dm312345',
         });
     });
     it('Adds default properties if config.yml is not fully specified', () => {
```
AlexNi245, line 79: This test is supposed to test, if the config reader adds proper default values when notificationsChannels are not defined in the config.yml. 
The expected outcome would be that the notification channel would be an empty array.

Plase then add another test to ensure that an email related channel can be added using the config.yml

packages/backend/src/config/getDeliveryServiceProperties.ts
```diff
@@ -2,19 +2,29 @@ import { parse } from 'yaml';
...
     messageTTL: 0,
     //100Kb
     sizeLimit: 100000,
-    notificationChannel: [],
-    smtpHost: '',
-    smtpPort: 0,
-    smtpEmail: '',
-    smtpUsername: '',
-    smtpPassword: '',
+    notificationChannel: [
```
AlexNi245, line 16: Per default the NotificationChannels are just an empty array

packages/backend/src/notifications.ts
```diff
@@ -91,13 +93,16 @@ export default () => {
...
                 });
             } else {
-                // Adding a user's notification channel to the database
-                await req.app.locals.db.addUsersNotificationChannel(account, {
-                    type: notificationChannelType,
-                    config: {
-                        recipientValue: recipientValue,
-                    },
-                });
+                // add new notification channel & send OTP for verification
```
AlexNi245, line 96: I would only pass the DB object once to the function instead then accessing the method from within the functions.

packages/backend/src/persistance/getDatabase.ts
```diff
@@ -90,6 +94,9 @@ export async function getDatabase(_redis?: Redis): Promise<IDatabase> {
         // Global Notification
         getGlobalNotification: Notification.getGlobalNotification(redis),
         setGlobalNotification: Notification.setGlobalNotification(redis),
+        // Otp
```
AlexNi245, line 97: Please add a little description what OTP actually is

packages/backend/src/persistance/otp/setOtp.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export function setOtp(redis: Redis) {
+    return async (
+        ensName: string,
+        otp: string,
+        channelType: NotificationChannelType,
+        generatedAt: Date,
+    ) => {
+        // Get existing OTP records from Redis
+        const existingOtp = await redis.get(
```
AlexNi245, line 14: I wonder if it wouldn't be better to have the OTP for a certain channel scoped in within the key. 
So for example when creating an OTP for Email the key would be RedisPrefix.Otp + (await getIdEnsName(redis)(ensName)) =EMAIL

That would lead to the system only have a valid OTP over time and reduce the complexity of both setting and retrieving the OTP from the DB

packages/lib/delivery/src/Notification.ts
```diff
@@ -0,0 +1,92 @@
...
+            recipientValue: recipientValue,
+        },
+    });
+
+    // generate and save OTP
+    const otp = await saveOtp(notificationChannelType, ensName, setOtp);
+
+    // Filter notification channels to send OTP
+    const notificationChannels = await getUsersNotificationChannels(ensName);
+    const filteredNotificationChannels = notificationChannels.filter(
```
AlexNi245, line 65: Isn't this function always throwing since the channels are initially empty

packages/lib/delivery/src/Notification.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    if (!filteredNotificationChannels.length) {
+        throw new Error('Invalid notification channel');
+    }
+
+    /**
+     * This is a mandatory property in sendNotification method so constructed
+     * from and to both with same ensname
+     */
+    const deliveryInformation: DeliveryInformation = {
```
AlexNi245, line 77: The reason deliveryInformation is part because it used to be needed by the first template. Now sendNotification can handle more than just one template. I think it would then make sense to change the API from deliveryInformation to generic content. object that can be set depending on what the template needs.

packages/lib/delivery/src/notifications/Notifications.test.ts
```diff
@@ -1,39 +1,53 @@
...
+            getDeliveryServiceProperties().notificationChannel;
 
-        const channel2 = {
+        const channel1 = {
             type: NotificationChannelType.EMAIL,
-            config: {},
+            config: {
+                recipientValue: 'bob@gmail.com',
+                isEnabled: true,
+                isVerified: false,
```
AlexNi245, line 33: Can you add a test that ensures that no mail gets ended when an account is not verified?

packages/backend/src/config/getDeliveryServiceProperties.test.ts
Bhupesh-mfsi: ok

packages/lib/delivery/src/notifications/getEmailTemplate.ts
```diff
@@ -5,11 +5,14 @@ import {
...
 import { OTP_EMAIL_SUBJECT, OTP_EMAIL_TEMPLATE } from './templates/otp';
-import { generateOtp } from './generateOtp';
+
+// OTP expiration time in seconds
+const OTP_EXPIRY_DURATION = 600; // 10 minutes
 
 // to fetch subject & template of email based on notification type
 export const fetchEmailSubjectAndTemplate = (
     notificationType: NotificationType,
+    mailContent: string,
```
AlexNi245, line 15: As mentioned above mailContant and deliveryInformation are related values to the corresponding template. Please use an generic values argument that supplies the selected template with all values needed.

packages/lib/delivery/src/notifications/templates/otp.ts
```diff
@@ -1,8 +1,173 @@
...
+                margin: 0 auto;
+                margin-top: 50px;
+                text-align: center;
+                font-weight: 500;
+                color: #ffffff;
+              "
+            >
+              Need help? Ask at
+              <a
+                href="mailto:decentralized.messaging@gmail.com"
```
AlexNi245, line 137: Supply as value

AlexNi245 (changes requested):
see comments

Bhupesh-mfsi:
> see comments

ok

packages/lib/delivery/src/notifications/broker/NotificationBroker.ts
```diff
@@ -27,25 +28,65 @@ export const _setupNotficationBroker = (
...
                 const deliveryServiceNotificationChannel =
                     supportedChannels.find((c) => c.type === channel.type);
-                //User specified a channel that is not supported.
-                //This should be prevented by refusing any schema that allows to provide a channel that is not supported
+                //User specified a channel that is not supported
                 if (!deliveryServiceNotificationChannel) {
                     throw new Error(
                         `Channel type ${channel.type} is not supported`,
                     );
+                } else if (
```
AlexNi245, line 41: please use if instead of else if here

packages/backend/src/notifications.ts
```diff
@@ -4,9 +4,17 @@ import { normalizeEnsName } from '@dm3-org/dm3-lib-profile';
...
+import {
+    ChannelNotSupportedError,
+    DeliveryServiceProperties,
+    addNewNotificationChannel,
+} from '@dm3-org/dm3-lib-delivery';
+import { IDatabase } from './persistance/getDatabase';
 
 // Exporting a function that returns an Express router
-export default () => {
+export default (
```
AlexNi245, line 15: why not passing the DS properties object directly ?

packages/backend/src/index.ts
```diff
@@ -70,7 +70,7 @@ global.logger = winston.createLogger({
     app.use('/storage', Storage());
     app.use('/auth', Auth());
     app.use('/delivery', Delivery());
-    app.use('/notifications', Notifications());
+    app.use('/notifications', Notifications(getDeliveryServiceProperties));
```
AlexNi245, line 73: DeliveryServiceproperties are already initialized in Line 63 use the value of that var to pass it through the function instead of providing a function

packages/backend/src/notifications.ts
Bhupesh-mfsi: I thought if needed to pass path to config file then function would be useful.

Bhupesh-mfsi:
> Please use the deliveryServiceValue instead of a function as an arugument for Notifications

Done

Bhupesh-mfsi:
> Please use the deliveryServiceValue instead of a function as an arugument for Notifications

Done

AlexNi245 (approved):

PR #756, opened 2024-02-09
748 resend otp

Bhupesh-mfsi:


packages/backend/src/notifications.ts
```diff
@@ -66,6 +70,58 @@ export default (deliveryServiceProperties: DeliveryServiceProperties) => {
...
+            // Fetch global notification data of user from database
+            const globalNotification =
+                await req.app.locals.db.getGlobalNotification(account);
+
+            // if global notification is turned off
+            if (!globalNotification.isEnabled) {
+                res.sendStatus(400).json({
+                    error: 'Global notifications is off',
+                });
+            } else {
```
AlexNi245, line 102: Don't use else, instead return in line 99 to exit the function

packages/lib/delivery/src/Notification.ts
```diff
@@ -90,3 +93,98 @@ const getOtpContentForNotificationChannel = (
             );
     }
 };
+
+// method to resend otp
+export const resendOtp = async (
```
AlexNi245, line 98: why is it called resend and not send otp ?

Bhupesh-mfsi: Yes, can be send OTP that's more better

PR #763, opened 2024-02-13
753 verify otp

Bhupesh-mfsi:


packages/lib/delivery/src/Notification.ts
```diff
@@ -161,6 +165,63 @@ export const sendOtp = async (
...
+// method to resend otp
+export const verifyOtp = async (
+    ensName: string,
+    notificationChannelType: NotificationChannelType,
+    otpToValidate: string,
+    dsNotificationChannels: NotificationChannel[],
+    db: any,
+) => {
+    // check if channel is supported or not
+    const channelUsed = dsNotificationChannels.filter(
```
AlexNi245, line 177: You might use .some() instead of filter() which gives returns you an boolean  indication if the value is part of the collection.

PR #764, opened 2024-02-14
enable & disable notification channel

Bhupesh-mfsi:


packages/backend/src/validation/notification/enableOrDisableChannelValidation.ts
```diff
@@ -0,0 +1,42 @@
+import { validateNotificationChannelType } from './notificationChannelValidation';
+
+// validates enable/disable notification channel data
+export const enableOrDisableNotificationChannelData = (
```
AlexNi245, line 4: please rename this function to validateToggleNotificationChannel

packages/backend/src/notifications.ts
```diff
@@ -182,6 +184,61 @@ export default (deliveryServiceProperties: DeliveryServiceProperties) => {
...
+
+            // if global notification is turned off
+            if (!globalNotification.isEnabled) {
+                return res.status(400).json({
+                    error: 'Global notifications is off',
+                });
+            }
+
+            // enable or disable notification
+            await enableOrDisableNotificationChannel(
```
AlexNi245, line 221: i think toggleNotificationChanel would be a more concise name

packages/lib/delivery/src/Notification.ts
```diff
@@ -222,6 +222,54 @@ export const verifyOtp = async (
...
+    );
+
+    if (!channelToUpdate.length) {
+        throw Error(
+            `${notificationChannelType} notification channel is not configured`,
+        );
+    }
+
+    // check if notification channel is alaready enabled/disabled
+    checkAlreadyEnabledOrDisabled(
```
AlexNi245, line 259: I think we don't need to throw an error here. If the user intends to activate the channel that is already active we can simply return 200 because his intent is fullfiled anyways

AlexNi245 (changes requested):
see comments

packages/lib/delivery/src/Notification.ts
Bhupesh-mfsi: ok

PR #790, opened 2024-02-22
add delivery-service workspace

### LINKED ISSUE #778: DS1 - Create own package for the DeliveryService
At the moment the DeliveryService is part of the overall dm3-backend workspace. It should become an independent package as it is supposed to be operated without the other backend components

- [x] Create a new workspace dm3-delivery-service
- [x] Move the express app that is currently located in delivery.ts into that workspace
- [x] rpc-proxy.ts contains a method called DM3_SUBMIT_MESSAGE move this method to the new workspace aswell
- [x] Integrate the new workspace to the build pipeline

malteish:
closes #778


package.json
```diff
@@ -1,5 +1,6 @@
 {
     "name": "dm3",
+    "private": true,
     "workspaces": {
```
AlexNi245, line 4: What is the reason to make it private? I think that'll lead to issues when the package is about to be published to npm

packages/delivery-service/src/index.ts
```diff
@@ -0,0 +1,104 @@
...
+     */
+    app.use('/profile', Profile()); // remove in backend
+    app.use('/delivery', Delivery()); // remove in backend
+    app.use(
+        '/notifications',
+        Notifications(app.locals.deliveryServiceProperties),
+    ); // remove in backend
+    app.use(logError);
+    app.use(errorHandler);
+    //@ts-ignore
```
AlexNi245, line 83: Why is ts-ignore used here?

packages/delivery-service/src/persistence/getIdEnsName.ts
```diff
@@ -0,0 +1,33 @@
...
+                RedisPrefix.Session + 'alias:' + normalizeEnsName(ensName),
+            )) ?? ensName,
+        );
+
+        return lowerEnsName === ensName ? ensName : resolveAlias(lowerEnsName);
+    };
+    return resolveAlias;
+}
+
+export function getAliasChain(redis: Redis) {
```
AlexNi245, line 17: This function can be removed

packages/delivery-service/src/persistence/session/setSession.ts
```diff
@@ -0,0 +1,28 @@
...
+            throw Error('Invalid session');
+        }
+        await redis.set(
+            RedisPrefix.Session + (await getIdEnsName(redis)(ensName)),
+            stringify(session),
+        );
+    };
+}
+
+export function setAliasSession(redis: Redis) {
```
AlexNi245, line 21: Can be removed aswell

AlexNi245 (changes requested):
see comments

package.json
malteish: [This is how a monorepo should be set up.](https://classic.yarnpkg.com/lang/en/docs/workspaces/), including the "private" setting. It prevents the monorepo itself being published by mistake, which it never should. Instead, the individual workspaces should be published, which is still possible.

packages/delivery-service/src/index.ts
malteish: Not sure, removing it now.

malteish: These [functions](https://github.com/dm3-org/dm3/blob/54347b4af7156bf5f34e5120d1eef747c0d692f4/packages/delivery-service/src/messaging.ts#L20) expect `app` to implement the `express` and `WithLocals` interfaces. They use `app.locals.[...]`. Any tips on how to explicitly check for these interfaces without bloating too much?

packages/delivery-service/src/persistence/getIdEnsName.ts
malteish: done

packages/delivery-service/src/persistence/session/setSession.ts
malteish: done

packages/delivery-service/src/index.ts
AlexNi245: The withLocals interface attempts to specify the structure of the req.app.locals object that otherwise would be empty. The main advantage here is that the compiler would know the function signatures of the local vats, especially the IDatebase var. 
To tell the compiler that you've assigned all the mandatory locals file between 58 and 64 use 
```app as express.Express & { locals: WithLocals["locals"] })```

Comment (malteish) [reply packages/delivery-service/src/index.ts]: thx

File: packages/backend/src/persistence/getDatabase.ts [hunk 5]
```diff
@@ -126,90 +83,20 @@ export async function getDatabase(
         getNumberOfConverations: Storage.getNumberOfConversations(prisma),
         //Storage Toggle Hide Conversation
         toggleHideConversation: Storage.toggleHideConversation(prisma),
-        //Get the user db migration status
```
AlexNi245, line 129: If the migration should be kept the methods related to the old storage and the migration should be kept

AlexNi245 (changes requested):
Pr looks good, however before merging this we should clarify how we want to handle migrations. 
I also think we should not merge directly to develop before we have not made a new release of the current develop branch.

malteish:
> Pr looks good, however before merging this we should clarify how we want to handle migrations. I also think we should not merge directly to develop before we have not made a new release of the current develop branch.

I agree, but think deploying the new storage to production would be a bad idea. Doing both changes at once (separation of backend and delivery service as well as heavily altering the storage format) seems clever.

PR #802, opened 2024-02-29
Widget fe optimization

Bhupesh-mfsi:


.vscode/settings.json
```diff
@@ -5,7 +5,7 @@
         "editor.formatOnSave": true
     },
     "[typescript]": {
-        "editor.defaultFormatter": "esbenp.prettier-vscode",
+        "editor.defaultFormatter": "vscode.typescript-language-features",
```
AlexNi245, line 8: why do we need to change the formatter?

packages/messenger-widget/src/components/ConfigureProfile/ConfigureProfile.tsx
```diff
@@ -119,16 +123,24 @@ export function ConfigureDM3Profile() {
 
     // changes network on naming service change
     const changeNetwork = (serviceName: string) => {
-        const chainId = fetchChainIdFromServiceName(serviceName);
-        if (chainId && chainId !== chain?.id) {
+        const chainId = fetchChainIdFromServiceName(
+            serviceName,
+            state.dm3Configuration.chainId,
+        );
+        if (chainId && chainId !== chainId) {
```
AlexNi245, line 130: Can this condition ever be true?

packages/messenger-widget/src/hooks/auth/useAuth.ts
```diff
@@ -4,6 +4,7 @@ import {
     normalizeEnsName,
 } from '@dm3-org/dm3-lib-profile';
 import { useContext, useEffect, useMemo, useState } from 'react';
+// @ts-ignore
```
AlexNi245, line 7: what is the reason for the ts-ignore at the wagmi imports?

packages/messenger-widget/src/contexts/DM3Configuration.ts
```diff
@@ -0,0 +1,39 @@
+import { log } from '@dm3-org/dm3-lib-shared';
+import {
+    DM3ConfigurationStateActions,
+    DM3ConfigurationStateType,
+} from '../utils/enum-type-utils';
+import { DM3Configuration } from '../interfaces/config';
+
+export function dm3ConfigurationReducer(
```
AlexNi245, line 8: Create a hook similar to useConversations or useMessage instead of using the legacy state

Bhupesh-mfsi: oh yes, that makes sense

packages/messenger-widget/src/hooks/auth/useAuth.ts
Bhupesh-mfsi: As per the wagmi documentation, a declaration for import wagmi needs to be created in the latest version. But when that declaration is made the imports from wagmi gives error. So its set to typescript ignored. But it works on execution.

packages/messenger-widget/src/components/ConfigureProfile/ConfigureProfile.tsx
Bhupesh-mfsi: ohh, same variable names made this confusion. Will fix this

packages/messenger-widget/src/views/Home/Home.tsx
```diff
@@ -29,87 +26,83 @@ import { WebSocketContextProvider } from '../../context/WebSocketContext';
...
-                            .walletConnectProjectId as string,
-                        chains,
-                    }),
-                    walletConnectWallet({
-                        projectId: props.config
-                            .walletConnectProjectId as string,
-                        chains,
-                    }),
-                ],
+                appName: 'DM3 app',
```
AlexNi245, line 54: I think the build is failing because the arguments here are not matching

packages/messenger-widget/src/hooks/conversation/useConversation.tsx
```diff
@@ -1,18 +1,20 @@
...
 import { hydrateContract } from './hydrateContact';
+import { fetchPendingConversations } from '../../adapters/messages';
+import { normalizeEnsName } from '@dm3-org/dm3-lib-profile';
+import { Config } from '../../interfaces/config';
+import { TLDContext } from '../../context/TLDContext';
+import { useDm3Configuration } from '../configuration/useDM3Configuration';
 
 export const useConversation = (config: Config) => {
     const mainnetProvider = useMainnetProvider();
+    const { dm3Configuration } = useDm3Configuration();
```
AlexNi245, line 17: I still think the configuration will always be empty since the invocation of the hook loads the default state. i think the hook has to be exported similarly to MessageContext to make it globally accessible

AlexNi245 (changes requested):
see comments

packages/messenger-widget/src/components/DM3/DM3.tsx
```diff
@@ -36,6 +39,9 @@ function DM3(props: Dm3Props) {
         if (childElement && childElement.parentElement) {
             childElement.parentElement.classList.add('h-100');
         }
+
+        // sets the DM3 confguration provided from props
+        setDm3Configuration(props.dm3Configuration);
```
Bhupesh-mfsi, line 44: @AlexNi245 Here the configuration set from the props

PR #807, opened 2024-03-04
add flag to indicate storageMigration

AlexNi245:


packages/backend/src/storage.test.ts
```diff
@@ -824,6 +843,36 @@ describe('Storage', () => {
...
+                .send();
+
+            expect(preMigrationStatus).toBe(false);
+
+            const { status } = await request(app)
+                .post(`/new/bob.eth/migrationStatus`)
+                .set({
+                    authorization: `Bearer ${token}`,
+                })
+                .send();
```
malteish, line 862: 1. I would prefer `startMigration` and `migrationStatus` or something similar in addition to `post` and `get` in order to make the difference more visible
2. Do we have any large accounts that would make finishing the migration while the call is open difficult? Then we would need a third status, e.g. `inProgress` or something similar

packages/backend/src/storage.ts
```diff
@@ -216,10 +214,30 @@ export default () => {
...
+            return res.json(status);
+        } catch (e) {
+            next(e);
+        }
+    });
+
+    router.post('/new/:ensName/migrationStatus', async (req, res, next) => {
+        try {
+            const ensName = normalizeEnsName(req.params.ensName);
+            await db.setUserDbMigrated(ensName);
```
malteish, line 230: Where does the actual migration happen?

malteish (changes requested):
This does add a flag, but does not migrate the data.
Why do we need a flag? We could say: If user exists in database, they were migrated. If not, check if they exist in old storage before creating an entry.

packages/backend/src/storage.ts
AlexNi245: Since the user data is encrypted the client has to to the migration. The flow would look like this

1. If old storage exists and not has been migrated
2. Get old Storage
3. Decrypt old Storage
4. Run Migration
5. Set the Migrated flag to true so the next time the app is opened the migration will not be triggered again.

malteish: Ok, that makes perfect sense.

PR #825, opened 2024-03-11
Contact UI responsive

Bhupesh-mfsi:


packages/messenger-widget/src/components/Chat/Chat.tsx
```diff
@@ -78,7 +77,7 @@ export function Chat(props: HideFunctionProps) {
             id="chat-msgs"
             className={'chat-msgs width-fill '
                 .concat(
-                    selectedContact
+                    selectedContact && screenWidth >= 800
```
AlexNi245, line 80: Please make 800 a global constant so we could adjust I later without going through every file

packages/messenger-widget/src/components/Contacts/Contacts.css
```diff
@@ -109,3 +109,50 @@
...
+        padding: 0.5rem 0.5rem 0.5rem 2rem !important;
+    }
+
+    .contact-details-container:nth-child(even) {
+        background: var(--alternate-contact-background-color) !important;
+    }
+
+    .contact-details-container:hover {
+        border-radius: 0px;
+        padding-left: 2rem !important;
```
AlexNi245, line 130: Why do we need to use important on every single entry?

AlexNi245 (changes requested):
see comments

packages/messenger-widget/src/components/Contacts/Contacts.css
Bhupesh-mfsi: Its because any other CSS don't override it. We have container queries as well as normal CSS so they don't override it.

AlexNi245 (approved):

PR #838, opened 2024-03-15
Resp add conv modal

Bhupesh-mfsi:
**Modifications done :-**
1. Added responsive UI for Add conversation modal.
2. Fixed ENS name bug instead of GNO name .
3. Fixed bug of Configure profile box still visible when profile is pubished.
4. Responsive UI for default contact enabled screen.
5. Added profile screen visibility for default contact enabled screen(which was not implemented before).

### DIFF
--- a/packages/messenger-widget/src/utils/ens-utils.ts
+++ b/packages/messenger-widget/src/utils/ens-utils.ts
@@ -95,14 +95,26 @@ export const onClose = (
     dispatch: React.Dispatch<Actions>,
     setSelectedContact: Function,
     screenWidth: number,
+    showContacts: boolean,
 ) => {
     if (screenWidth && screenWidth > MOBILE_SCREEN_WIDTH) {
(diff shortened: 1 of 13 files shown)


packages/messenger-widget/src/utils/ens-utils.ts
```diff
@@ -95,14 +95,26 @@ export const onClose = (
     dispatch: React.Dispatch<Actions>,
     setSelectedContact: Function,
     screenWidth: number,
+    showContacts: boolean,
 ) => {
     if (screenWidth && screenWidth > MOBILE_SCREEN_WIDTH) {
```
AlexNi245, line 100: remove else and add screenWidth && screenWidth > MOBILE_SCREEN_WIDTH to the if condition of showcontacts

AlexNi245 (changes requested):
see comments

AlexNi245 (approved):

Issue #827, opened 2024-03-12
DS 1.5


malteish:
based on the [separation of backend and ds](https://github.com/dm3-org/dm3/pull/790), extract common components for both to use:

- [ ]  auth 
- [ ]  utils

malteish:
done

PR #840, opened 2024-03-19
Feature/extract auth

malteish:
see #827


malteish:
solved by #873 and #870

PR #870, opened 2024-04-02
Feature/extract auth only

malteish:
see #827


packages/backend/src/index.ts
```diff
@@ -66,7 +66,10 @@ global.logger = winston.createLogger({
     });
     app.use('/profile', Profile());
     app.use('/storage', Storage(app.locals.db));
-    app.use('/auth', Auth());
+    app.use(
+        '/auth',
+        Auth(app.locals.db.getSession as any, app.locals.db.setSession as any),
```
AlexNi245, line 71: instead of casting the function to any you should be able to use the interface from iDatabase here

AlexNi245 (approved):

packages/backend/src/index.ts
malteish: sadly iDatabase by now diverged between backend and delivery service, which means I can not use it here. Not sure yet how to treat that.

PR #878, opened 2024-04-03
Feature/separation of backend and delivery service

malteish:
see #778 and #827


packages/backend/src/profile.test.ts
```diff
@@ -99,26 +126,20 @@ describe('Profile', () => {
                 signature,
             };
 
-            const { status } = await request(app)
+            const response = await request(app)
                 .post(`/${wallet.address}`)
                 .send(signedUserProfile);
 
+            // log status message
+            console.log(response.info);
```
AlexNi245, line 134: Do we need this log here?  Otherwise we can remove it to keep the logs clean

packages/delivery-service/src/index.ts
```diff
@@ -0,0 +1,104 @@
...
+    const io = new Server(server, {
+        cors: {
+            origin: '*',
+            methods: 'GET,HEAD,PUT,PATCH,POST,DELETE',
+            preflightContinue: false,
+            optionsSuccessStatus: 204,
+        },
+    });
+
+    app.use(logRequest);
```
AlexNi245, line 62: what is this ?

packages/delivery-service/src/index.ts
```diff
@@ -0,0 +1,104 @@
...
+    /**
+     *     needed
+     */
+    app.use('/profile', Profile(db, web3Provider, io));
+    app.use('/delivery', Delivery(web3Provider, db));
+    app.use(
+        '/notifications',
+        Notifications(deliveryServiceProperties, db, web3Provider),
+    );
+    app.use(logError);
```
AlexNi245, line 77: whats the difference between error handler and logError ?

packages/delivery-service/src/index.ts
```diff
@@ -0,0 +1,104 @@
...
+            deliveryServiceProperties,
+            io,
+            web3Provider,
+            db,
+        ),
+    );
+})();
+
+// TODO include standalone web app
+app.use(express.static(path.join(__dirname, '../../web/build')));
```
AlexNi245, line 95: I know it was part of the old app but this is probably the worst way a client application can be served. Before the refactoring, it worked because the CI started two backend containers but with a different command. 
Is that still the case, except the container would be delivery-service now?

packages/backend/src/profile.test.ts
```diff
@@ -1,85 +1,112 @@
...
 
-            const { status } = await request(app)
+            const response = await request(app)
                 .get('/0x99C19AB10b9EC8aC6fcda9586E81f6B73a298870')
+                .set({
+                    authorization: `Bearer ${mockToken}`,
+                })
                 .send();
 
+            console.log(response.error);
```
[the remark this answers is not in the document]
malteish: true, removed.

packages/backend/src/profile.test.ts
malteish: true, removed

packages/delivery-service/src/index.ts
malteish: This logs some info about each request: https://github.com/dm3-org/dm3/blob/890d8f9b56431a47f70faa4c9e013e03421243ff/packages/lib/server-side/src/utils.ts#L96

packages/delivery-service/src/index.ts
malteish: `logError` logs the error while `errorHandler` alters the applications response and terminates the routine. `logError` is used much more frequently than `errorHandler` in our codebase (44 times vs 2 times), so separating them seems a good idea. But I guess we could make `errorHandler` call `logError`. Wdyt?

packages/delivery-service/src/index.ts
malteish: I have no idea what was served here, but assume the delivery service does not need to serve this static content anymore. I removed it. Can you tell me more about the background?

packages/delivery-service/migrations/20240229170017_/migration.sql
```diff
@@ -0,0 +1,36 @@
+-- CreateTable
+CREATE TABLE "EncryptedMessage" (
```
AlexNi245, line 2: Isn't that supposed to be part of the BE. If the prisma schema has moved there wen don't need to keep the migration in the delivery-service dir

packages/delivery-service/migrations/migration_lock.toml
```diff
@@ -0,0 +1,3 @@
+# Please do not edit this file manually
+# It should be added in your version-control system (i.e. Git)
```
AlexNi245, line 2: do we have prisma at the ds ?  if not remove

packages/delivery-service/src/persistence/pending/index.ts
```diff
@@ -0,0 +1,5 @@
+import { addPending } from './addPending';
```
AlexNi245, line 1: I think the entire storage dir should no longer be part of ds since it has moved to BE

packages/delivery-service/src/validation/notification/enableOrDisableChannelValidation.ts
```diff
@@ -0,0 +1,42 @@
...
+            notificationChannelType,
+        );
+
+        // return error if channel type is not valid
+        if (!isChannelValid.isValid) {
+            return isChannelValid;
+        }
+
+        // throw error if isEnabled not found
+        if (isEnabled !== false && !isEnabled) {
```
AlexNi245, line 23: what is this ?

AlexNi245: this statement never will be true

AlexNi245 (changes requested):
Overall PR looks good, great stuff that'll move us forward. Besides some small requests, we're ready to merge it.

packages/delivery-service/migrations/20240229170017_/migration.sql
malteish: absolutely: fab74eba9a84c4d87707db43e4745d51c0c87a66

packages/delivery-service/src/persistence/pending/index.ts
malteish: yes: b97020fb0befefee621e00c0547d7f401f6f6864

packages/delivery-service/migrations/migration_lock.toml
malteish: removed: fab74eb

packages/delivery-service/src/validation/notification/enableOrDisableChannelValidation.ts
malteish: I don't know that this does and have found it [in the backend,](https://github.com/dm3-org/dm3/blob/b97020fb0befefee621e00c0547d7f401f6f6864/packages/backend/src/validation/notification/enableOrDisableChannelValidation.ts#L22) too.

malteish: removed from backend cfc0dfa

malteish: I can make this statement true manually in node:
```
> enOrDis.validateToggleNotificationChannel("EMAIL",null);
{ isValid: false, errorMessage: 'isEnabled value is missing' }
```

AlexNi245 (approved):

PR #912, opened 2024-05-02
removed global config from widget client

Bhupesh-mfsi:


packages/messenger-widget/src/components/RightHeader/NormalView.tsx
```diff
@@ -31,6 +31,7 @@ export function NormalView() {
             await getAvatarProfilePic(
                 mainnetProvider,
                 account?.ensName as string,
+                dm3Configuration.addressEnsSubdomain,
```
AlexNi245, line 34: Instead of passing the address via function args use the Cpntext directly from with the AvatarProfilePic component. Do this for all components.

AlexNi245 (changes requested):
see comments

AlexNi245 (approved):

PR #921, opened 2024-05-07
Bug delete session id on disconnect

AlexNi245:


packages/backend/src/ws/WebSocketManager.ts
```diff
@@ -0,0 +1,111 @@
...
+     * Adds a new connection to the connections map.
+     * @private
+     * @param {Socket} connection - The socket connection instance.
+     */
+    private async addConnection(connection: Socket) {
+        try {
+            const { account, token } = connection.handshake.auth;
+
+            const ensName = normalizeEnsName(account.ensName);
+            //Use the already existing function cechkToken to check weather the token matches the provided ensName
```
malteish, line 60: cechkToken -> checkToken, weather (as in rain) -> whether

packages/backend/src/ws/WebSocketManager.ts
```diff
@@ -0,0 +1,111 @@
...
+            const hasSession = await checkToken(
+                this.web3Provider,
+                this.db.getSession,
+                ensName,
+                token,
+            );
+            //retrive the session from the db
```
malteish, line 67: retri**e**ve

packages/backend/src/ws/WebSocketManager.ts
```diff
@@ -0,0 +1,111 @@
...
+                connection.disconnect();
+            }
+            //Get the old connections and add the new one
+            const oldConnections = this.connections.get(ensName) || [];
+            this.connections.set(ensName, [...oldConnections, connection]);
+            //Send the authorized event
+            connection.emit(AUTHORIZED);
+            console.log('connection established for ', ensName);
+            //When the socket disconnects we wan't them no longer in our connections List
+            connection.on('disconnect', () => {
```
malteish, line 82: did you check if anybody ever removes a connection if the client simply loses their internet connection? How does this happen? Is it after a timeout? If so, how long is that timeout?

malteish (changes requested):
This will be a pain to merge with the separation PR, but otherwise looks okay. Please check my comments.

packages/backend/src/ws/WebSocketManager.ts
```diff
@@ -0,0 +1,111 @@
...
+import { checkToken } from '@dm3-org/dm3-lib-delivery';
+import { IDatabase } from '../persistence/getDatabase';
+import { ethers } from 'ethers';
+import { IWebSocketManager } from '@dm3-org/dm3-lib-shared';
+
+export const UNAUTHORIZED = 'unauthorized';
+export const AUTHORIZED = 'authorized';
+
+export class WebSocketManager implements IWebSocketManager {
+    private readonly connections: Map<string, Socket[]> = new Map();
```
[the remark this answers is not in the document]
AlexNi245: Yes, what is read-only is the pointer to the Map.  So it can't be resigned after the object was constructed. 
[This article gives a good explanation of why it's advantageous to use readonly/final](https://stackoverflow.com/a/154510)

packages/backend/src/ws/WebSocketManager.ts
AlexNi245: disconnect will be called if the connection has been interrupted regardless the reason
Can be either called programmatically by the client, or more likely terminated on the networking layer (the client has been closed)

In both cases the disconnected event will be invoked

malteish (approved):
lgtm

PR #958, opened 2024-05-22
Webpush notification

Bhupesh-mfsi:
Webpush notification frontend & backend


packages/messenger-widget/README.md
```diff
@@ -44,6 +44,8 @@ Follow the below given steps :-
     REACT_APP_WALLET_CONNECT_PROJECT_ID=27b3e102adae76b4d4902a035da435e7
     REACT_APP_CHAIN_ID=11155111
     REACT_APP_RESOLVER_ADDR=0xae6646c22D8eE6479eE0a39Bf63B9bD9e57bAD9d
+    REACT_APP_GENOME_REGISTRY_ADDRESS=0x5dC881dDA4e4a8d312be3544AD13118D1a04Cb17
+    REACT_APP_PUBLIC_VAPID_KEY=BFCJLre0GeM6S-n4mkMX4SLZWlDR9qc8RsHyctsOPh_QDQkBuvCrMe9Rmq24736F-CJFp-3DkDWhp19X7mOJrEc
```
AlexNi245, line 48: what is the vapid key ?

packages/messenger-widget/src/worker.js
```diff
@@ -0,0 +1,15 @@
+import logo from './assets/images/dm3.svg';
+
+// service worker to detect any push notification is triggered
+self.addEventListener('push', (ev) => {
```
AlexNi245, line 4: is it possible to change the event name to dm3_register_push_notification to have the namespace clear?

AlexNi245 (changes requested):
see comments

packages/messenger-widget/README.md
Bhupesh-mfsi: Its a public key used to subscribe to push manager for the browser. The same key is used in backend as well. For more details please check this document https://thecodebarbarian.com/sending-web-push-notifications-from-node-js.html

packages/messenger-widget/src/worker.js
Bhupesh-mfsi: In the documentation, I can't see anywhere that the event name can be changed. So we have to remain stick with this.

PR #966, opened 2024-05-29
Dsfe 4   apply new auth to fe

AlexNi245:


packages/lib/server-side/src/auth.ts
```diff
@@ -98,6 +98,7 @@ export const Auth = (getSession, serverSecret: string) => {
             res.json(jwt);
         } catch (e) {
             next(e);
+            console.log();
```
AlexNi245, line 101: remove

Bhupesh-mfsi, line 101: It's a empty console, if not needed can be removed

Bhupesh-mfsi (approved):

PR #973, opened 2024-05-29
Add LRU cache to web3 provider 

AlexNi245:


malteish (commented):
Where does this come from and why do we trust it to work without unit tests?

AlexNi245:
> Where does this come from and why do we trust it to work without unit tests?

The cache is already used within Gashawk and tested there

malteish:
> > Where does this come from and why do we trust it to work without unit tests?
> 
> The cache is already used within Gashawk and tested there

Why not take the tests from there and using them here, too? Also, adding a reference to the gashawk repo in comments would be a good idea.

malteish (approved):
nice!

PR #983, opened 2024-06-05
add resolve config to root package.json

malteish:


malteish:
seems not to work and unsure about the benefit

PR #992, opened 2024-06-06
get rid of pending conversations

AlexNi245:


packages/backend/src/index.ts
```diff
@@ -43,14 +41,6 @@ winston.loggers.add('default', global.logger);
 
 (async () => {
     //TODO REPLACE WITH WS MANAGER
```
malteish, line 43:
```suggestion
```

malteish (changes requested):
lgtm, but let's remove the stale comment

PR #997, opened 2024-06-06
fixed EMAIL & PUSH notification issue

Bhupesh-mfsi:


packages/delivery-service/src/validation/notification/regexValidation.ts
```diff
@@ -17,7 +17,7 @@ export const checkRegexPattern = (
                 .match(EMAIL_REGEX);
             return patternCheck;
         case NotificationChannelType.PUSH:
-            return typeof recipientValue !== 'string' ? true : false;
+            return recipientValue.length > 300 ? true : false;
```
AlexNi245, line 20: what is this line doing ?

AlexNi245 (changes requested):
see comments

packages/delivery-service/src/validation/notification/regexValidation.ts
Bhupesh-mfsi: Th subscription value that comes from frontend is of type string and genrally its value is of minimum length 300, so that validation is set.

packages/lib/delivery/src/notifications/channels/Push.ts
```diff
@@ -3,21 +3,40 @@ import webpush from 'web-push';
...
+                    subject: serverConfig.vapidEmailId,
+                    publicKey: serverConfig.publicVapidKey,
+                    privateKey: serverConfig.privateVapidKey,
+                },
+            };
+
             const payload = JSON.stringify({
                 body: `You received a new message from ${config.notificationContent.from}.`,
                 title: 'New Message',
+                image: 'https://framerusercontent.com/images/6kPsfvAEqACFXgMOWU539mUqfB0.png',
```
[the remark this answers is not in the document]
Bhupesh-mfsi: The browser notification that user will see like this 
Screenshot from 2024-06-10 14-44-11
There is a logo of DM3, this URL is for that logo only. I took this URL from our website https://dm3.network/
This will be changed as we will decide in future.

packages/delivery-service/src/validation/notification/regexValidation.ts
AlexNi245: we cant use the length of the string to determine the notification channel that is just to opaque. Why do we not check if the string contains all the information needed to parse the channel

Bhupesh-mfsi: Okay, I will make the changes in validation

packages/delivery-service/src/validation/notification/regexValidation.ts
```diff
@@ -17,7 +17,13 @@ export const checkRegexPattern = (
                 .match(EMAIL_REGEX);
             return patternCheck;
         case NotificationChannelType.PUSH:
-            return typeof recipientValue !== 'string' ? true : false;
+            const data = JSON.parse(recipientValue);
```
AlexNi245, line 20: looks way better. Please get rid of the doubled negation to make it perfect.
return data.endpoint &&
                data.keys &&
                data.keys.p256dh &&
                data.keys.auth

AlexNi245 (approved):

AlexNi245 (approved):

PR #998, opened 2024-06-07
Generate dedicated docker images for every service and finish setup of environment "testing"

### LINKED ISSUE #955: Create dedicated docker images for each service
Backend, frontend and offchain-resolver all use the generalized `backend`-image, which is huge and too complex. 
We already create a dedicated image for the delivery service. Apply the same approach to the three services mentioned above to improve composability and resource consumption.

malteish:
closes #955 

- creates several docker images on push
- fixes some stuff that wasn't set up correctly in the repo, like missing dependencies.


docker/Dockerfile
```diff
@@ -1,7 +1,70 @@
...
+RUN mkdir /library_archives
+RUN for lib in `ls -I billboard-api`; do cd $lib; yarn pack --filename /library_archives/$lib.tgz; cd ..;  done
+
+## prepare dependency operations for all packages
+WORKDIR '/build/packages'
+# several steps are performed here for each package individually (but not for the libraries):
+# 1. extract the information which libraries are needed for this package and store it in required_libraries
+# 2. remove all dependencied referencing "workspace" in the monorepo from package.json
+# 3. update the package.json with the resolutions from the docker/resolutions.json file
+RUN for package in `ls -I lib`;\
```
AlexNi245, line 23: Can you add a tiny comment above each bash line to explain what it does? That would increase the readability a lot

docker/Dockerfile
```diff
@@ -1,7 +1,70 @@
-FROM --platform=linux/amd64 node:22-alpine
-WORKDIR /app
+FROM --platform=linux/amd64 node:22-alpine as build
```
AlexNi245, line 1: Isn't this file about to be removed?

packages/delivery-service/src/delivery.ts
```diff
@@ -1,7 +1,7 @@
 import { Acknoledgment, getMessages, schema } from '@dm3-org/dm3-lib-delivery';
 import { auth } from '@dm3-org/dm3-lib-server-side';
 import { validateSchema } from '@dm3-org/dm3-lib-shared';
-import { getConversationId } from '@dm3-org/dm3-lib-storage';
+import { getConversationId } from '@dm3-org/dm3-lib-delivery';
```
AlexNi245, line 4: why has this moved to delivery?

AlexNi245 (changes requested):

packages/delivery-service/src/delivery.ts
malteish: It was actually already in dm3-lib-delivery, and also in lib-storage. As the delivery service needs the function, but doesn't use lib storage, I pointed it to lib-delivery. 
Then, afaik, getConversationId in lib-storage was not used anymore, so I removed it.

docker/Dockerfile
malteish: true. removed it. 41838f3

packages/delivery-service/src/delivery.ts
malteish: image
Also, it is only used in ds atm

docker/Dockerfile
malteish: done 23e9b1a

PR #1004, opened 2024-06-10
Add test helper lib package

AlexNi245:


malteish (changes requested):
will this ever be used outside the messenger-widget? If not, we should move it into the messenger-widget package.

Resolutions looks good.

AlexNi245:
> will this ever be used outside the messenger-widget? If not, we should move it into the messenger-widget package.
> 
> Resolutions looks good.

Not yet but they could be handy in the future especially for BE and DS. So I'd keep them shared

AlexNi245:
> > will this ever be used outside the messenger-widget? If not, we should move it into the messenger-widget package.
> > Resolutions looks good.
> 
> Not yet but they could be handy especially for BE and DS. So I'd keep them shared

i.E messageFactory, UserProfile. They are easy to use and used on various occasions.

PR #1014, opened 2024-06-13
Feature/finalize jwt

### LINKED ISSUE #909: check jwt implementation
- [x] Would it be a good idea to use asymmetric keys in our jwt? Are they available both in backend and delivery service? Then do it.
- [x] issue jwt with challenge instead of keeping challenge ich db
- [ ] ~clean up db schema: remove session completely~ -> moved to #907 
- [x] do we need to [check what the ens resolves to](https://github.com/dm3-org/dm3/blob/fd39b0b9e32f0aae0994ab8a395b42d1ef0af458/packages/lib/delivery/src/Session.ts#L34) when checking the existing token? That doesn't actually answer any questions imo.
- [x] ~change auth message definition: why `Bearer` -> remove, make key shorter for efficiency.~ Seems to be standard, so keep it.
- [x] add `nbf` (not before) to token instead of manually verifying `iat`
- [x] check schema of auth token with schema tool instead of manually
- [x] decide for `user` or `account` in both tokens

malteish:
closes #909

### DIFF
--- a/packages/lib/delivery/src/Session.ts
+++ b/packages/lib/delivery/src/Session.ts
@@ -1,10 +1,22 @@
 import { ethers } from 'ethers';
 import { ProfileExtension, SignedUserProfile } from '@dm3-org/dm3-lib-profile';
-import { logDebug } from '@dm3-org/dm3-lib-shared';
+import { logDebug, validateSchema } from '@dm3-org/dm3-lib-shared';
 import { verify, decode } from 'jsonwebtoken';
 
 //1Year
 const TTL = 31536000000;
+
+const authJwtPayloadSchema = {
+    type: 'object',
+    properties: {
+        account: { type: 'string' },
+        iat: { type: 'number' },
+        exp: { type: 'number' },
+        nbf: { type: 'number' },
+    },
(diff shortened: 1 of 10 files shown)


packages/lib/delivery/src/Session.ts
```diff
@@ -57,48 +53,28 @@ export async function checkToken(
...
+        console.log('jwt payload', decode(token));
 
-        // check if expected fields are present
+        // check if payload is well formed
         if (
-            !('account' in jwtPayload) ||
-            !('iat' in jwtPayload) ||
-            !('exp' in jwtPayload) ||
-            !('nbf' in jwtPayload)
+            typeof jwtPayload == 'string' ||
```
AlexNi245, line 60: Checking for strict equality should always preferred to avoid unintended behavior.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness

AlexNi245 (changes requested):
see comments

packages/lib/delivery/src/Session.ts
malteish: right, fixed

PR #1042, opened 2024-06-21
1032 add backend pagination / refine api

AlexNi245:


packages/backend/migrations/20240621131510_/migration.sql
malteish, line 1: do we really want a different data type for this column compared to createdAt?

packages/backend/src/persistence/storage/postgres/getMessages.ts
```diff
@@ -1,17 +1,21 @@
...
-
-const PAGE_SIZE = 100;
 
 export const getMessages =
     (db: PrismaClient) =>
-    async (ensName: string, encryptedContactName: string, page: number) => {
+    async (
+        ensName: string,
+        encryptedContactName: string,
+        size: number,
```
malteish, line 8: I think `chunkSize`/`pageSize` and `chunkOffset`/`pageOffset` or something similar would be clearer. I hear `size`, I think kilobytes. I hear `offset`, I think data structure or something.

packages/backend/src/storage.ts
```diff
@@ -62,21 +66,38 @@ export default (
...
-            req.body;
+        const {
+            encryptedEnvelopContainer,
+            encryptedContactName,
+            messageId,
+            createdAt,
+        } = req.body;
 
-        if (!encryptedEnvelopContainer || !encryptedContactName || !messageId) {
+        if (
```
malteish, line 76: any reason not to use the schema validator tool here?

packages/backend/src/storage.ts
```diff
@@ -62,21 +66,38 @@ export default (
...
+            !createdAt
+        ) {
             res.status(400).send('invalid schema');
             return;
         }
 
         try {
             const ensName = normalizeEnsName(req.params.ensName);
+            //Since the message is fully encrypted, we cannot use the messageHash as an identifier.
+            //Instead we use the hash of the ensName and the messageId to havea unique identifier
```
malteish, line 89: havea -> have a
```suggestion
            //Instead we use the hash of the ensName and the messageId to have a unique identifier
```

packages/backend/src/storage.ts
```diff
@@ -187,7 +213,21 @@ export default (
     router.get('/new/:ensName/getConversations', async (req, res, next) => {
         try {
             const ensName = normalizeEnsName(req.params.ensName);
-            const conversations = await db.getConversationList(ensName);
+            const size =
+                parseInt(req.query.size as string) ||
```
malteish, line 217: what would happen on size=0? Would be a frontend error obviously, but what does the backend do?

malteish (changes requested):

packages/backend/src/storage.ts
AlexNi245: it would return an empty array as requested by the client. There might be situations where the client might calculate such a query. This then is not inherently wrong in my opinion

packages/backend/migrations/20240621131510_/migration.sql
AlexNi245: We can go with DateTime in both rows, but I'd use the Linux timestamp at the API since its more common to use than an iso-8601 string which Prisma uses to serialize a DateTime field.

packages/backend/src/storage.ts
AlexNi245: Until now we only used the schema validation tools for the lib function. However, I think it'll improve our code quality a lot if we use it on the backend as well.

I'll add schema generation similar to the lib package to the storage endpoints

malteish (approved):

PR #1061, opened 2024-06-26
use prisma migrate deploy instead of dev

### LINKED ISSUE #1056: prisma uses `migrate dev`, which should only be used in development
https://www.prisma.io/docs/orm/prisma-migrate/workflows/development-and-production

https://github.com/dm3-org/dm3/blob/develop/packages/backend/package.json#L31

recommendation: use `prisma migrate deploy` instead. If that is bad for local development, separate the dev and production commands.

malteish:
closes #1056

@AlexNi245 this is untested, can you checkout the branch and run it locally first please?

replaces #1060


malteish:
well, something doesn't work.

PR #1063, opened 2024-06-27
Contact pagination fe

Bhupesh-mfsi:


packages/messenger-widget/src/hooks/conversation/useConversation.tsx
```diff
@@ -209,12 +209,22 @@ export const useConversation = (config: DM3Configuration) => {
             DEFAULT_CONVERSATION_PAGE_SIZE,
             Math.floor(offset),
         );
+
+        let newContactAddedCount = 0;
```
AlexNi245, line 213: Please return conversation.length instead of counting it separatly

packages/messenger-widget/src/hooks/conversation/useConversation.tsx
```diff
@@ -209,12 +209,22 @@ export const useConversation = (config: DM3Configuration) => {
...
+
         conversations.forEach((conversation) => {
-            _addConversation(
-                conversation.contactEnsName,
-                conversation.isHidden,
-            );
+            // check if the contact is really added in the list or not
+            // if its added it will return contact object else nothing
+            if (
```
AlexNi245, line 218: don't use the if statement here either

packages/messenger-widget/src/hooks/conversation/useConversation.tsx
Bhupesh-mfsi: Here if somehow our own account address is found, then it is not added in contact list. But it will be counted in conversations.length. So to avoid that count is used

AlexNi245: The account cannot by part of the contact list at this part because it is not possible to add it in the first place

AlexNi245: So we can ignore it here

Bhupesh-mfsi: okay

AlexNi245 (approved):

PR #1107, opened 2024-07-17
fixed msg input field auto clear bug

Bhupesh-mfsi:


packages/messenger-widget/src/components/MessageInputBox/MessageInputBox.tsx
```diff
@@ -56,12 +59,27 @@ export function MessageInputBox() {
...
+     */
     useEffect(() => {
-        setMessageView({
-            actionType: MessageActionType.NONE,
-            messageData: undefined,
-        });
-        setMessage('');
+        if (
+            !contactSelected ||
+            contactSelected.contactDetails.account.ensName !==
```
AlexNi245, line 73: isn't that always false ?

AlexNi245: would it be cleaner to fix the selectedContact in useConversation instead of having it redundant here ?

Bhupesh-mfsi: But there also I need to have extra state to know whether contact selected is changed or just updatedAt property of contact is changed due to new message

AlexNi245 (approved):

Issue #1091, opened 2024-07-15
Remove redis from bakend


malteish:
#907 revealed that redis is only ever used for `Session` creation in the backend, where `Session` actually refers to an account. As the prisma model already defines an `Account`, we decided to remove redis and use the existing prisma `Account` model instead. 

These fields exist in `Session`, if they are not crossed our they should be migrated to `Account`:

- [ ]    ~~account: string;~~  -> id
- [ ]     ~~signedUserProfile: SignedUserProfile;~~ -> must be read from resolver
- [ ]     ~~token: string;~~ -> deprecated since jwt
- [ ]     publicMessageHeadUri?: string;
- [ ]     createdAt: number;
- [ ]     ~~socketId?: string;~~ -> deprecated since socket manager
- [ ]     ~~challenge?: string;~~ deprecated since jwt
- [ ]     profileExtension: ProfileExtension;
- [ ]     ~~spamFilterRules?: unknown;~~ -> spam filter does not exist yet, add field if necessary

malteish:
migration plan: 
1. add createdAt field to Account model, populate with now as default
2. replace all `getSession` (old) or `getAccount` (new) calls in backend with `checkAccountExists` that uses prisma as data source
3. test on testing
4. add manual migration that adds accounts in postgres for all sessions in redis
5. test on staging

malteish:
In the backend, redis is only used to store the "Session". Todo:

- create a table "account" in postgres that can be queried to know if an ens name has signed up with the backend yet
- populate this table from redis through some scripts
- change getSession to use this new postgres datasource
- remove redis from backend

Moved this info here and closed #1089

malteish:
Estimate: 9 days (3 remaining)

PR #1109, opened 2024-07-17
Use prisma accounts in backend

malteish:
see #1091 

- store accounts in prisma instead of redis in backend
- add scripts and notes for manual migration of existing accounts (formerly sessions) from redis to prisma
- separate profile management in backend and delivery service, as the overlap now is negligible
- extend account table to include created at
- clean up libraries, improving structure. server-side and delivery saw most changes and have been detangled significantly.
- we used to have 2 files named Auth.ts and auth.ts. Imo Auth.ts handles authentication and is currently still called Auth.ts, while auth.ts handles authorization and has been renamed to authorize.ts. This starts a cleanup process that will be continued later, but this PR is already large enough.
- move socketAuth to delivery-service (from server-side), because it is used only there


packages/backend/src/persistence/storage/postgres/doesAccountExist.ts
```diff
@@ -0,0 +1,17 @@
...
+export const doesAccountExist =
+    (db: PrismaClient) => async (ensName: string) => {
+        //Check if account exists
+        const account = await db.account.findFirst({
+            where: {
+                id: ensName,
+            },
+        });
+        //If account exists, return true
+        if (account) {
```
AlexNi245, line 12: return !!account returns a bool indicating if the account is falsey or not

packages/lib/server-side/src/utils.ts
```diff
@@ -16,7 +16,7 @@ export async function auth(
     res: Response,
     next: NextFunction,
     ensName: string,
-    db: IAccountDatabase,
+    doesAccountExist: (ensName: string) => Promise<boolean>,
```
AlexNi245, line 19: A name like hasAccount would be more concise. getAccount, setAccount, hasAccount

malteish: good idea!

packages/backend/src/persistence/storage/postgres/doesAccountExist.ts
```diff
@@ -0,0 +1,17 @@
+import { PrismaClient } from '@prisma/client';
+
+export const doesAccountExist =
```
AlexNi245, line 3: what is the benefit of using doesAccountExist over !!getAccount()

malteish: this is used in lib shared or something, too, and delivery-service and backend now have different return types on getAccount, which causes trouble. Introducing this extra function solves that.

packages/lib/server-side/src/utils.ts
malteish: done

packages/backend/src/persistence/storage/postgres/doesAccountExist.ts
malteish: done

malteish:
todo: fix tests by doing the todos in these files:
https://github.com/dm3-org/dm3/blob/e14667ab8e6fe2f2ab55f6d9b56f7ac0eb2cc75f/packages/backend/src/persistence/session/setAccount.test.ts

https://github.com/dm3-org/dm3/blob/e14667ab8e6fe2f2ab55f6d9b56f7ac0eb2cc75f/packages/backend/src/persistence/session/setAccount.test.ts

@AlexNi245 maybe you can help with that? I'm having trouble finding the right words to talk to the coding gods.

packages/backend/src/persistence/account/setAccount.test.ts
```diff
@@ -0,0 +1,37 @@
...
+            publicEncryptionKey: '',
+            publicSigningKey: '',
+            deliveryServices: [],
+        };
+
+        const priorSetAccount = await db.getAccount(USER_NAME);
+
+        //User has no account yet
+        expect(priorSetAccount).toBe(null);
+        await db.setAccount(USER_NAME);
```
AlexNi245, line 31: you might have to pass a session as a second FN arg

packages/backend/src/persistence/storage/postgres/setAccount.ts
```diff
@@ -0,0 +1,6 @@
+import { PrismaClient } from '@prisma/client';
+import { getOrCreateAccount } from './utils/getOrCreateAccount';
+
+export const setAccount = (db: PrismaClient) => async (ensName: string) => {
+    return getOrCreateAccount(db, ensName);
```
AlexNi245, line 5: You've to add the actual account as an arg. At the moment only empty accounts can be created with getAccount

malteish:
The interface IAccountDatabase is used in these locations:
1. server-side/auth -> needs getAccount
2. server-side/profile -> needs getAccount, setAccount
  - only used in ds, backend has it's own implementation 
4. server-side/socketAuth -> needs hasAccount, getAccount (only to resolve address) and setAccount (for old ws manager)
  - getAccount is used in socketAuth to get the account address, but account.account mostly contains these patterns: "<address>.addr.dm3.eth"
  - setAccount is used for wsManager, 
  - only used in ds -> can be moved to lib/delivery and use the IDSDatabase interface
  - 
  
Replacing getAccount with hasAccount in auth is enough to use auth for both backend and ds. Profile and SocketAuth are only used in ds anyway.

malteish:
db.getAccount is overridden in ds index to resolve ens for address and lookup with address.

packages/backend/src/persistence/storage/postgres/setAccount.ts
malteish: The new setAccount function does not need more arguments.

packages/backend/src/storage.ts
```diff
@@ -32,7 +32,15 @@ export default (
             next: NextFunction,
             ensName: string,
         ) => {
-            auth(req, res, next, ensName, db, web3Provider, serverSecret);
+            authorize(
```
AlexNi245, line 35: I think the best name for the function would be authMiddleware. Since it is a express middleware intercepting the request https://expressjs.com/en/guide/using-middleware.html

packages/delivery-service/src/persistence/account/hasAccount.ts
```diff
@@ -0,0 +1,13 @@
...
+import { getIdEnsName } from '../getIdEnsName';
+
+export function hasAccount(redis: Redis) {
+    return async (ensName: string) => {
+        let account = await redis.get(
+            RedisPrefix.Account + (await getIdEnsName(redis)(ensName)),
+        );
+
+        // return true if account exists, false otherwise
+        return account ? true : false;
```
AlexNi245, line 11: return !! account

packages/lib/server-side/src/authorize.ts
```diff
@@ -0,0 +1,97 @@
...
+    next: NextFunction,
+    ensName: string,
+    hasAccount: (ensName: string) => Promise<boolean>,
+    web3Provider: ethers.providers.JsonRpcProvider,
+    serverSecret: string,
+) {
+    const normalizedEnsName = normalizeEnsName(ensName);
+    const authHeader = req.headers['authorization'];
+    const token = authHeader && authHeader.split(' ')[1];
+    //TODO resolve addr for ens name
```
AlexNi245, line 31: remove TODO as its not longer up to date

AlexNi245 (approved):

packages/delivery-service/src/persistence/account/hasAccount.ts
malteish: changed it accordingly

malteish:
I tested the full migration process on testing, and it worked nicely. Here are my notes from the test:

- [ ] no backend accounts or sessions
- [ ] - sign up with account 1 -> 2 signatures 
- [ ] - sign up with account 2 -> 2 signatures
- [ ] - send message between account 1 and account 2
- [ ] - backend has 2 accounts 
- [ ] - sign up with account 3 -> 2 signatures
- [ ] - backend has 3 sessions
- [ ] - accounts have no createdAt
- [ ] - deploy new version
- [ ] - backend has accounts created at 'now', and sessions
- [ ] - sign in with account 1 -> 1 signature
- [ ] - extract sessions
- [ ] - overwrite accounts
- [ ] - the account's createdAt now corresponds to the session's createdAt
- [ ] - sign in with account 2 -> 1 signature
- [ ] - sign in with account 3 -> 1 signature
- [ ] - send messages between accounts -> works
- [ ] - sign up with account 4 -> 2 signatures
- [ ] - 4 accounts, 3 sessions

We are ready to to this on staging and production next week.

packages/backend/src/storage.ts
malteish: I think "auth" in "authMiddleware" would still not clearly state what the function does, while "Middleware" doesn't add much value even though it makes up two thirds of the name. How about "authorizationMiddleware" or just leaving it at "authorize"?

PR #1112, opened 2024-07-18
Testing

AlexNi245:


malteish:
- add variable for persistence directory
- reset testing state automatically
- reset staging state manually
- use prisma deploy instead of dev
- remove winston from backend
- fix typo acknoledgment
- acknowledge message hashed instead of timestamps

todo:

- [ ] check why session in getDatabase not removed
- [ ] remove pendi

ng message schema

PR #1119, opened 2024-07-19
set node docker container to 3.18

malteish:
node:22-alpine-3.19 was released yesterday and
causes our builds to fail. Reverting to last
working version 3.18.


malteish:
closed for #1121

PR #1131, opened 2024-07-30
Old account signin issue

Bhupesh-mfsi:


packages/lib/server-side/src/auth.ts
```diff
@@ -96,8 +96,13 @@ export const Auth = (getAccount, serverSecret: string) => {
 
             res.json(jwt);
         } catch (e) {
-            next(e);
             console.error('unable to create new session token ', e);
+            if (e instanceof Error && e.message === 'Signature invalid') {
```
AlexNi245, line 100: i think we could return every error with code 400

     return res.status(400).json({
                    error: e.message,
                });

packages/messenger-widget/src/hooks/server-side/ServerSideConnector.ts
```diff
@@ -100,11 +103,21 @@ export abstract class ServerSideConnector extends JwtInterceptor {
         );
 
         //Todo move to lib
-        const { data: newToken } = await axios.post(url, {
+        const {
```
AlexNi245, line 106: this might not work without catching the error

AlexNi245 (changes requested):
see comments

packages/messenger-widget/src/hooks/server-side/ServerSideConnector.ts
```diff
@@ -89,23 +93,39 @@ export abstract class ServerSideConnector extends JwtInterceptor {
...
+                data: newToken,
+                status,
+                data: error,
+            } = await axios.post(url, {
+                signature,
+                challenge,
+            });
 
-        return newToken;
+            return newToken;
```
AlexNi245, line 119: please add some comments why this step might occur

AlexNi245 (approved):

packages/messenger-widget/src/hooks/server-side/ServerSideConnector.ts
Bhupesh-mfsi: added

PR #1132, opened 2024-08-01
Halt delivery msg fix

Bhupesh-mfsi:


packages/messenger-widget/src/hooks/haltDelivery/useHaltDelivery.ts
```diff
@@ -145,8 +155,13 @@ export const useHaltDelivery = () => {
 
             await submitEnvelopsToReceiversDs(dispatchableEnvelops);
 
+            // delete the halted messages with messageId
             dispatchableEnvelops.map((envelop) => {
-                clearHaltedMessages(envelop.haltedEnvelopId, envelop.aliasName);
+                // fetch the messageId from message signature
+                if (envelop.envelop.metadata?.signature) {
+                    const msgId = msgMap.get(envelop.envelop.message.signature);
```
AlexNi245, line 162: I think it would cause problems this way because the signatures from a dispatchable envelope is different from its original envelope. After all, the deliveryinformation is also part of the signature.

AlexNi245, line 162: Isn't haltedEnevelpeId defined in 138 the same as msgId in 162?

Bhupesh-mfsi: No, those are different

Bhupesh-mfsi: We use message signature and not the envelop signature, so I don't think it should mess up because message signature should remain same.

packages/messenger-widget/src/hooks/haltDelivery/useHaltDelivery.ts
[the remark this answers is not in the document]
Bhupesh-mfsi: Yes, this makes sense and will work

packages/lib/storage/src/new/types.ts
```diff
@@ -7,7 +7,7 @@ export interface StorageAPI {
         pageSize: number,
         offset: number,
     ) => Promise<StorageEnvelopContainer[]>;
-    getHaltedMessages: () => Promise<StorageEnvelopContainer[]>;
+    getHaltedMessages: () => Promise<any[]>;
```
AlexNi245, line 10: Please provide the proper type instead of just returning any. If there is no type yet add one to the types.ts file

PR #1141, opened 2024-08-12
Remove redis from backend

malteish:
- removes temporary files
- removes redis from backend
- removes the redis container from all environments (testing, staging, production)
- renames Session to Account in delivery service


malteish:
Do not merge this before I completed #1091, or we will lose data!

PR #1142, opened 2024-08-12
update and sync prisma packages

### LINKED ISSUE #1122: prisma version mismatch
When building the backend, I see this warning: 
`warn Versions of prisma@5.10.1 and @prisma/client@4.16.2 don't match.`

The mismatch comes from [package.json](https://github.com/dm3-org/dm3/blob/develop/packages/backend/package.json).

I believe we should fix it.

malteish:
closes #1122


malteish:
can't fix the broken build

malteish:
closing to favor #1143

PR #1143, opened 2024-08-12
Sync prisma versions

### LINKED ISSUE #1122: prisma version mismatch

malteish:
closes #1122


packages/backend/package.json
```diff
@@ -13,14 +13,13 @@
     "@dm3-org/dm3-lib-shared": "workspace:^",
     "@dm3-org/dm3-lib-storage": "workspace:^",
     "@dm3-org/dm3-lib-test-helper": "workspace:^",
-    "@prisma/client": "4.16.2",
+    "@prisma/client": "latest",
```
AlexNi245, line 16: I think we should pin the latest LTS of prisma instead of always using the latest version

AlexNi245 (changes requested):
see comment

packages/backend/package.json
malteish: I don't find any information on LTS versions, so I pinned the latest major version. Minor versions will be updated automatically, which is the same approach as we have with all other packages.

PR #1157, opened 2024-08-28
Lukso poc

AlexNi245:


packages/lib/server-side/src/web3Provider/getLuksoProvider.ts
```diff
@@ -0,0 +1,17 @@
+import { ethers } from 'ethers';
+
+export async function getLuksoProvider(
+    env: NodeJS.ProcessEnv,
```
karladler, line 4: access process.env directly? no argument needed

packages/lib/server-side/src/web3Provider/getLuksoProvider.ts
```diff
@@ -0,0 +1,17 @@
...
+): Promise<ethers.providers.JsonRpcProvider> {
+    const readKey = (keyName: string) => {
```
karladler, line 6: maybe use generic utils function like this

```typescript
export const checkEnvVars = (keys: string[]): void => {
  keys.forEach((varname) => {
    if (typeof process.env[varname] !== 'string') {
      throw new Error(`Environment variable '${varname}' is not set!`);
    }
  });
};


// call with

checkEnvVars(['LUKSO_RPC'])
```

packages/lib/smart-account/src/KeyStore/IKeyStore.ts
```diff
@@ -0,0 +1,28 @@
+import { SignedUserProfile } from '@dm3-org/dm3-lib-profile';
+
+//Struct that represents a Dm3 key store
+export interface Dm3KeyStore {
```
karladler:
```suggestion
export interface IDm3KeyStore {
```

packages/lib/smart-account/src/KeyStore/IKeyStore.ts
```diff
@@ -0,0 +1,28 @@
...
+    [signerAddress: string]: {
+        signerPublicKey: string;
+        encryptedProfileKeys?: string;
+    };
+}
+
+export interface IKeyStoreService {
+    //Method to publish the Dm3 user profile. Ideally this should be ENS or any other place that can be access using a resolver
```
karladler: Using block comment style would allow the IDE to render a nice message on hover. Looks not so nice in the code, but might be useful for SDK users.

```suggestion
     /**
     * Method to publish the Dm3 user profile. Ideally this should be ENS or any other place that can be access using a resolver
     */
```

Screenshot 2024-11-12 at 19 04 07

packages/lib/smart-account/src/lukso/ERC725JsonCoder.test.ts
```diff
@@ -0,0 +1,26 @@
...
+            signature: '0x101',
+        };
+
+        const wallet = ethers.Wallet.createRandom();
+        const encoded = new ERC725JsonCoder().encodeDm3Profile(
+            wallet.address,
+            signedUserProfile,
+        );
+
+        console.log('enc', encoded);
```
karladler: looks like there is some actual test missing?

packages/lib/smart-account/src/lukso/ERC725JsonCoder.ts
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    public async decodeDm3Profile(
+        upAddress: string,
+    ): Promise<SignedUserProfile | undefined> {
+        const encodedUserProfile = await this.erc725.getData({
+            keyName: 'DM3UserProfile:<address>',
+            dynamicKeyParts: upAddress,
+        });
+        if (encodedUserProfile.value === null) {
```
karladler: we could get rid of the following `as Data[]` `as String` using a typeguard like

```ts
function isStringArray(value: unknown): value is string[] {
    return Array.isArray(value) && value.every((item) => typeof item === 'string');
}

// then instead of just checking for null:
 if (!isStringArray(encodedUserProfile.value)) {
    return undefined;
}
```

packages/lib/smart-account/src/lukso/ERC725JsonCoder.ts
```diff
@@ -0,0 +1,159 @@
...
+                ),
+            };
+            return acc;
+        }, {} as Dm3KeyStore);
+
+        return ks;
+    }
+}
+//"0x89197c814f5df4249c180000c73abaa79d9d562d07f09d8135b72b047908bbe6"
+//"0x89197c814f5df4249c1800007870c5b8bc9572a8001c3f96f7ff59961b23500d"
```
karladler: ?

packages/lib/smart-account/src/lukso/LuksoIndexer.ts
```diff
@@ -0,0 +1,60 @@
...
+        this.client = client;
+    }
+
+    public async resolveAddress(address: string): Promise<string | undefined> {
+        const query = gql`
+            query MyQuery {
+                Profile(where: { id: { _ilike: "${address}" } }) {
+                    fullName
+                }
+            }
```
karladler: i think that gql queries shall be built more like this:

```ts
            query getProfile($address: string!) {
                Profile(where: { id: { _ilike: $address } }) {
                    fullName
                }
            }
```
also not sure about the API here, but there is probably also `_eq` to get a quicker result?

karladler: maybe we could also use gql codegen to streamline the process of generating gql functions.  
Might also get us fully E2E typescript support
https://github.com/dotansimha/graphql-code-generator

packages/lib/smart-account/src/lukso/LuksoKeyStore.ts
```diff
@@ -0,0 +1,97 @@
...
+                a,
+                keyStore[a].signerPublicKey,
+                keyStore[a].encryptedProfileKeys,
+            );
+        });
+        //TODO find diff and post only modified keys
+        const keys = encodeDataReturn.map((a) => a.keys[0]);
+        const values = encodeDataReturn.map((a) => a.values[0]);
+
+        await this.upContract.setDataBatch(keys, values);
```
karladler, line 60: await seems to be redundant here.

packages/messenger-widget/src/components/AddConversation/AddConversation.tsx
```diff
@@ -94,12 +94,23 @@ export default function AddConversation() {
...
         setShowError(false);
         setTldName(e.target.value);
-        if (!ethers.utils.isValidName(e.target.value)) {
+
+        if (
+            !(
+                ethers.utils.isValidName(e.target.value) ||
+                isLuksoName(e.target.value)
+            )
+        ) {
```
karladler: same logic, but might be easier to read:
```suggestion
        if (!ethers.utils.isValidName(e.target.value) && !isLuksoName(e.target.value)) {
```

packages/messenger-widget/src/components/AddConversation/AddConversation.tsx
```diff
@@ -94,12 +94,23 @@ export default function AddConversation() {
...
             setErrorMsg('Invalid address or ENS name');
             setShowError(true);
         }
     };
 
+    const isLuksoName = (input: string): boolean => {
```
karladler: should be defined before its called

packages/messenger-widget/src/components/SignIn/SignIn.tsx
karladler, line 1: actually touch the UI components in this PR? If yes, maybe hide that behind some feature flag?

packages/messenger-widget/src/components/SignIn/SignIn.tsx
```diff
@@ -97,6 +107,26 @@ export function SignIn() {
...
 
+                        {isLukso && (
+                            <LoginButton
+                                text="Sign In with Universal Profile"
+                                onClick={() => luksoSignIn()}
+                                buttonState={ButtonState.Ideal}
+                            />
+                        )}
+
+                        <LoginButton
```
karladler, line 118: test button is necessary?

packages/messenger-widget/src/hooks/auth/AccountConnector.ts
```diff
@@ -67,17 +67,21 @@ export const AccountConnector = (
...
-        const isProfileValid = await checkUserProfile(
-            mainnetProvider,
-            onChainProfile,
-            address,
-        );
-
-        if (!isProfileValid) {
-            throw Error('Profile signature is invalid');
-        }
+        //First occurance users logs in with onchain profile
```
karladler: any todos here?

packages/messenger-widget/src/hooks/messages/useMessage.tsx
```diff
@@ -351,6 +351,13 @@ export const useMessage = () => {
         return { isSuccess: true };
     };
 
+    //Just for testing purposes
+    //TODO cleanup and find better solution
+    const isLuksoName = (input: string): boolean => {
```
karladler: can be put into some common utils function since its duplicated here

packages/messenger-widget/src/hooks/topLevelAlias/nameService/UniversalProfile.ts
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    //e.g. 0x1234.addr.dm3.eth => alexcv#d7ab.up
+    async resolveAliasToTLD(
+        ensName: string,
+        foreignTldName?: string,
+    ): Promise<string> {
+        const [address] = ensName.split('.');
+        const aliasName = await this.luksoIndexer.resolveAddress(address);
+        return aliasName! + '.up';
```
karladler, line 42: I would aim to use always template literals where possible.

```suggestion
        return `${aliasName}.up`;
```

karladler (commented):
LGTM, please check comments

packages/lib/smart-account/src/lukso/ERC725JsonCoder.test.ts
AlexNi245: We should add the test when before we release

packages/lib/smart-account/src/lukso/LuksoIndexer.ts
AlexNi245: I changed the query name to a more expressive name. I think the reason for not using _eq is to support checksum addresses as well.

packages/messenger-widget/src/components/AddConversation/AddConversation.tsx
AlexNi245: Good catch!

packages/messenger-widget/src/hooks/auth/AccountConnector.ts
AlexNi245: No it has been resolved already

packages/lib/smart-account/src/lukso/LuksoKeyStore.ts
AlexNi245: I like that it is more descriptive than just knowing that the FN returns a Promise anyways. But we can change it ofc if its a concern for you

PR #1162, opened 2024-09-02
Add metrics api

malteish:
First step of enabling metrics collection.

Changes:
- introduce 2 new values to delivery service properties: metrics collection interval duration and metrics retention duration
- count messages received by ds, including size. 
- count notifications sent by ds
- make counted values available at the `/metrics` endpoint: https://testing.dm3.network/ds/metrics
- censor the current collection interval's values to improve privacy


packages/backend/package.json
```diff
@@ -31,7 +31,7 @@
     "start": "yarn prisma-init && node  ./dist/index.js",
     "start-inspect": "node --inspect=0.0.0.0:9229  ./dist/index.js",
     "test": "yarn run before:tests && DATABASE_URL='postgresql://prisma:prisma@localhost:5433/tests?schema=public' yarn jest --coverage --runInBand --transformIgnorePatterns 'node_modules/(?!(dm3-lib-\\w*)/)'",
-    "build": "yarn prisma generate && yarn tsc && cp ./config.yml ./dist/config.yml | true",
+    "build": "yarn prisma generate && yarn tsc | true",
```
AlexNi245, line 34: Can the container access the config.yml after it has been deployed via the CI ? I think we used to copy the file to dist due to CI-related issues. But not sure if they have been resolved with the new setup

AlexNi245 (approved):
lgtm

packages/backend/package.json
malteish: this is a remnant of the past: the backend doesn't use config.yml at all.

PR #1186, opened 2024-09-24
Metrics docs and disabling

malteish:
add some docs

Manually implement the disabling of metrics collection. While the original idea that redis will expire data immediately if retention is set to 0 is correct, it has some drawbacks:
- not explicit
- performs unnecessary write operations
- generates logs
-> better to manually prevent the writes in the first place.

### DIFF
--- a/packages/delivery-service/src/persistence/metrics/setMetrics.ts
+++ b/packages/delivery-service/src/persistence/metrics/setMetrics.ts
@@ -7,6 +7,11 @@ export function countMessage(redis: Redis) {
         messageSizeBytes: number,
         deliveryServiceProperties: DeliveryServiceProperties,
     ) => {
+        if (deliveryServiceProperties.metricsRetentionDurationInSeconds === 0) {
+            // Metrics are disabled
+            return;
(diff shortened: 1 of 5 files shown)


packages/delivery-service/src/persistence/metrics/setMetrics.ts
```diff
@@ -7,6 +7,11 @@ export function countMessage(redis: Redis) {
         messageSizeBytes: number,
         deliveryServiceProperties: DeliveryServiceProperties,
     ) => {
+        if (deliveryServiceProperties.metricsRetentionDurationInSeconds === 0) {
```
AlexNi245, line 10: What happens if the number is negative? Does the schema validation prevent this? Otherwise, it might make sense to use deliveryServiceProperties.metricsRetentionDurationInSeconds <= 0 instead of deliveryServiceProperties.metricsRetentionDurationInSeconds === 0

malteish: good hint, changed to <=

PR #1204, opened 2024-11-21
Add basic Messages and Conversation API

AlexNi245:
- Test setup for Dm3 SDK
- Basic API to add manage conversations
- Basic API to add new messages
- Display messages via messageRendere

### DIFF
--- a/packages/js-sdk/package.json
+++ b/packages/js-sdk/package.json
@@ -1,4 +1,3 @@
-
 {
     "name": "@dm3-org/dm3-js-sdk",
     "license": "BSD-2-Clause",
--- a/packages/js-sdk/package.json
+++ b/packages/js-sdk/package.json
@@ -1,3 +1,4 @@
+
 {
     "name": "@dm3-org/dm3-js-sdk",
     "license": "BSD-2-Clause",
--- a/packages/js-sdk/package.json
+++ b/packages/js-sdk/package.json
@@ -27,6 +27,7 @@
         "@babel/preset-env": "^7.19.4",
         "@babel/preset-typescript": "^7.18.6",
         "@types/supertest": "^2.0.12",
+        "axios-mock-adapter": "^2.1.0",
         "babel-cli": "^6.26.0",
         "babel-jest": "^29.2.2",
         "babel-preset-env": "^1.7.0",
--- a/packages/js-sdk/src/Dm3Sdk.test.ts
+++ b/packages/js-sdk/src/Dm3Sdk.test.ts
@@ -9,9 +9,11 @@ import { Dm3Sdk, Dm3SdkConfig } from './Dm3Sdk';
 
 import MockAdapter from 'axios-mock-adapter';
 import { normalizeEnsName } from '@dm3-org/dm3-lib-profile';
+import { ITLDResolver } from './tld/nameService/ITLDResolver';
 
 describe('Dm3Sdk', () => {
     let alice: MockedUserProfile;
+    let bob: MockedUserProfile;
 
     //Axios mock to mock the http requests
     let axiosMock;
@@ -22,9 +24,12 @@ describe('Dm3Sdk', () => {
             'alice.up',
             ['test.io'],
         );
+        bob = await mockUserProfile(ethers.Wallet.createRandom(), 'bob.up', [
+            'test.io',
+        ]);
     });
 
-    it('test', async () => {
+    it('can add a conversaton to the contact list', async () => {
         axiosMock = new MockAdapter(axios);
         //Mock BackendConnector HttpRequests
         //Mock profileExistsOnDeliveryService
@@ -52,15 +57,24 @@ describe('Dm3Sdk', () => {
             )
             .reply(200, 'mock-challenge');
 
+        const mockTldResolver = {
+            resolveTLDtoAlias: async () =>
+                `${normalizeEnsName(bob.address)}.addr.test`,
+            resolveAliasToTLD: async () => 'bob.eth',
+        } as unknown as ITLDResolver;
+
         const mockConfig: Dm3SdkConfig = {
             mainnetProvider: {} as ethers.providers.JsonRpcProvider,
-            storageApi: {} as StorageAPI,
+            storageApi: {
+                addConversation: async () => {},
+            } as unknown as StorageAPI,
             nonce: '1',
             defaultDeliveryService: 'test.io',
             addressEnsSubdomain: '.addr.test',
             userEnsSubdomain: '.user.test',
             resolverBackendUrl: 'resolver.io',
             backendUrl: 'http://localhost:4060',
+            _tld: mockTldResolver,
         };
 
         const dm3 = await new Dm3Sdk(mockConfig).login({
@@ -69,10 +83,10 @@ describe('Dm3Sdk', () => {
             accountAddress: alice.address,
         });
 
-        console.log(dm3.conversations.conversations);
-
-        /*      await dm3.conversations.addConversation('karl.eth');
-        const c = dm3.conversations.conversations;
-        const karl = c[0]; */
+        await dm3.conversations.addConversation('bob.eth');
+        const c = dm3.conversations.list;
+        console.log(c);
+        expect(c.length).toBe(1);
+        expect(c[0].contact.name).toBe('bob.eth');
     });
 });
--- a/packages/js-sdk/src/Dm3Sdk.test.ts
+++ b/packages/js-sdk/src/Dm3Sdk.test.ts
@@ -42,7 +42,9 @@ describe('Dm3Sdk', () => {
             ethers.Wallet.createRandom(),
             'http://localhost:3000',
         );
+    });
 
+    beforeAll(() => {
         axiosMock = new MockAdapter(axios);
 
         //Mock BackendConnector HttpRequests
@@ -102,9 +104,8 @@ describe('Dm3Sdk', () => {
             });
 
             await dm3.conversations.addConversation('bob.eth');
-            const c = dm3.conversations.list;
-            expect(c.length).toBe(1);
-            expect(c[0].contact.name).toBe('bob.eth');
+            expect(dm3.conversations.list.length).toBe(1);
+            expect(dm3.conversations.list[0].contact.name).toBe('bob.eth');
         });
         it('can multiple conversations to the contact list', async () => {
             const mockTldResolver = {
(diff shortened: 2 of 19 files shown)


packages/js-sdk/package.json
```diff
@@ -1,3 +1,4 @@
+
```
karladler:
```suggestion
```

packages/js-sdk/src/Dm3Sdk.test.ts
```diff
@@ -1,34 +1,278 @@
...
+    let alice: MockedUserProfile;
+    let bob: MockedUserProfile;
+    let karl: MockedUserProfile;
+
+    //Axios mock to mock the http requests
+    let axiosMock;
+
+    let deliveryService: MockDeliveryServiceProfile;
 
     beforeEach(async () => {
```
karladler, line 27: would it make more sense to put the axios mocks in `beforeAll` ?

packages/js-sdk/src/Dm3Sdk.test.ts
```diff
@@ -1,34 +1,278 @@
...
+            .onPost(
+                `http://localhost:4060/auth/${normalizeEnsName(
+                    alice.address,
+                )}.addr.test`,
+            )
+            .reply(200);
+    });
+
+    describe('conversations', () => {
+        it('can add a conversation to the contact list', async () => {
```
karladler, line 76: just for future tests, consider using `it should ...` pattern so that tests form a sentence in combination with `describe` e.g. "Testing conversations -> should add a conversation to the contact list"
And then the expect checks exactly for the "should" part.

packages/js-sdk/src/Dm3Sdk.test.ts
```diff
@@ -1,34 +1,278 @@
...
+                storageApi: {
+                    addConversation: async () => {},
+                } as unknown as StorageAPI,
+                nonce: '1',
+                defaultDeliveryService: 'test.io',
+                addressEnsSubdomain: '.addr.test',
+                userEnsSubdomain: '.user.test',
+                resolverBackendUrl: 'resolver.io',
+                backendUrl: 'http://localhost:4060',
+                _tld: mockTldResolver,
```
karladler, line 94: if the only purpose of `_tld` in config is to mock the function, we could also probably use jest mocking instead.
just to keep in mind for future refactorings.

https://jestjs.io/docs/mock-function-api#mockfnmockimplementationfn:~:text=jest.mocked(SomeClass).mockImplementation(()%20%3D%3E%20%7B%0A%20%20return%20%7B%0A%20%20%20%20method%3A%20mockMethod%2C%0A%20%20%7D%3B%0A%7D)%3B

packages/js-sdk/src/Dm3Sdk.test.ts
```diff
@@ -1,34 +1,278 @@
...
+                profileKeys: alice.profileKeys,
+                profile: alice.signedUserProfile,
+                accountAddress: alice.address,
+            });
+
+            await dm3.conversations.addConversation('bob.eth');
+            await dm3.conversations.addConversation('karl.eth');
+            const c = dm3.conversations.list;
+            dm3;
+            console.log(c);
```
karladler:
```suggestion
```

packages/js-sdk/src/Dm3Sdk.test.ts
```diff
@@ -1,34 +1,278 @@
...
+            const dm3 = await new Dm3Sdk(mockConfig).login({
+                profileKeys: alice.profileKeys,
+                profile: alice.signedUserProfile,
+                accountAddress: alice.address,
+            });
+
+            await dm3.conversations.addConversation('bob.eth');
+            await dm3.conversations.addConversation('karl.eth');
+            const c = dm3.conversations.list;
+            dm3;
```
karladler: ?

packages/js-sdk/src/Dm3Sdk.test.ts
```diff
@@ -1,34 +1,278 @@
...
+                profileKeys: alice.profileKeys,
+                profile: alice.signedUserProfile,
+                accountAddress: alice.address,
+            });
+
+            await dm3.conversations.addConversation('bob.eth');
+            await dm3.conversations.addConversation('bob.eth');
+            const c = dm3.conversations.list;
+            expect(c.length).toBe(1);
+            expect(c[0].contact.name).toBe('bob.eth');
```
karladler: should suffice without additional variable?

```suggestion
            expect(dm3.conversations.list.length).toBe(1);
            expect(dm3.conversations.list[0].contact.name).toBe('bob.eth');
```

packages/js-sdk/src/Dm3Sdk.test.ts
AlexNi245: That is a good take i normally follow. When I continue with the API implementation I'll use a cleaner API naming convention.

packages/js-sdk/src/Dm3Sdk.test.ts
AlexNi245: Thats an intresting though. I don't see a reason why tld should be overwritten. However given the object might enhance flexibility. 
In that case jest mock could be an interesting option to reduce complexity

packages/js-sdk/src/message/Messages.ts
```diff
@@ -13,62 +12,68 @@ import {
...
 
-    private messages: MessageStorage = {};
+    private readonly _messages: MessageModel[];
+
+    private readonly senderAccount: Account;
+    private readonly senderProfileKeys: ProfileKeys;
+    private readonly receiver: Contact;
 
-    constructor(storageApi: StorageAPI, conversations: Conversations) {
+    constructor(
```
karladler, line 32: Would be probably better to use names props here, but good for now.

AlexNi245: I agree. Lets do a final refiment once the SDK is stable enough

packages/js-sdk/src/message/Messages.ts
```diff
@@ -13,62 +12,68 @@ import {
...
+    ) {
         this.storageApi = storageApi;
         this.conversations = conversations;
+        this.senderAccount = senderAccount;
+        this.senderProfileKeys = senderProfileKeys;
+        this.receiver = receiver;
+        this._messages = [];
+    }
+
+    public list() {
```
karladler: could it be also a getter?

packages/js-sdk/src/message/Messages.ts
```diff
@@ -13,62 +12,68 @@ import {
...
+    public async sendMessage(msg: string) {
+        const messageWithoutSig: Omit<Message, 'signature'> = {
+            message: msg,
+            attachments: [],
+            metadata: {
+                referenceMessageHash: undefined,
+                type: 'NEW',
+                to: this.receiver.account.ensName,
+                from: this.senderAccount.ensName,
+                timestamp: new Date().getTime(),
```
karladler: just shorter:

```suggestion
                timestamp: Date.now(),
```

packages/js-sdk/src/message/Messages.ts
AlexNi245: yes that is indeed better

packages/js-sdk/src/message/Messages.ts
```diff
@@ -225,21 +230,21 @@ export class Messages {
...
-    };
+    }
     //TODO migrate to real lukso name service
     isLuksoName = (input: string): boolean => {
         const regex = /^[a-zA-Z0-9]+#[a-zA-Z0-9]{4}\.up$/;
         return regex.test(input);
     };
 
-    private checkIfEnvelopAreInSizeLimit = async (
+    private async checkIfEnvelopAreInSizeLimit(
```
karladler: async here necessary?

packages/js-sdk/src/message/renderer/messageTypes/renderDelete.ts
```diff
@@ -0,0 +1,62 @@
...
+                MessageActionType.DELETE,
+        )
+        .map((message) => {
+            if (
+                toBeDeletedByMessageHash.includes(
+                    message.envelop.metadata?.messageHash,
+                )
+            ) {
+                return {
+                    ...message,
```
karladler, line 30: looks like we might wrap this in a some kind of merge function or even use lodash merge function here.

packages/js-sdk/src/message/renderer/messageTypes/renderDuplicates.ts
```diff
@@ -0,0 +1,14 @@
+import { MessageModel } from '../../types';
+
+export const renderDuplicates = (messages: MessageModel[]) => {
```
karladler, line 3: I would rename it into `removeDuplicates`

packages/js-sdk/src/message/renderer/renderMessage.ts
```diff
@@ -0,0 +1,36 @@
...
+ */
+export const renderMessage = (messages: MessageModel[]) => {
+    const withReadReceived = renderReadReceived(messages);
+    const withReadOpened = renderReadOpened(withReadReceived);
+    const withDeletes = renderDelete(withReadOpened);
+    const withReactions = renderReactions(withDeletes);
+    const withReply = renderReply(withReactions);
+
+    const withoutEdited = renderEdit(withReply);
+    //Sort the messages by timestamp DESC to show them in the right order
```
karladler: is that done somewhere else yet?

packages/js-sdk/src/message/renderer/messageTypes/renderReactions.ts
```diff
@@ -0,0 +1,45 @@
...
+
+                return {
+                    ...message,
+                    reactions: _reactions,
+                };
+            })
+            //Get get rid of the reactions, as they are now part of the message
+            .filter(
+                (message) =>
+                    message.envelop.message.metadata.type !== 'REACTION',
```
karladler, line 44: can we use `MessageActionType.REACT` here?

packages/js-sdk/src/tld/Tld.ts
```diff
@@ -48,6 +48,17 @@ export class Tld {
...
+    isResolverForTldName(ensName: string): Promise<boolean> {
+        //Since its the root resolver its always capable of resolving
+        return Promise.resolve(true);
+    }
+    isResolverForAliasName(
+        ensName: string,
+        foreignTldName?: string,
+    ): Promise<boolean> {
+        //Since its the root resolver its always capable of resolving
+        return Promise.resolve(true);
```
karladler, line 60: seems to be a redundant, what's is that used for?

packages/js-sdk/src/Dm3Sdk.test.ts
AlexNi245: The axios mocks have the address as a dependency, which is generated beforeEach the new test. i think its better to create a new wallet with beforeEach for every test to provide a clean environment

packages/js-sdk/src/message/renderer/messageTypes/renderDelete.ts
AlexNi245: what do you mean by merge function? Are you referring to the fact that properties nested into the object are subject to change?

packages/js-sdk/src/message/renderer/messageTypes/renderDuplicates.ts
AlexNi245: It has been moved to renderEdit. Will remove the misleading comment

packages/js-sdk/src/tld/Tld.ts
AlexNi245: The interface isResolverForAliasName / isResolverForTldName are required by the ITLDResolver interface. It returns always true because the tld.ts Is the RootResolver and can always be used to retrieve a name because it knows every delegate.

